PR #1, opened 2015-10-09
Add travis files

squirrelo:


squirrelo:
This is good to merge, but we need to figure out a test database setup if we want to fully test things. Can we have this pull the AG info over and setup a database that way?

mortonjt:
I don't have merge rights atm.

squirrelo:
Right, new repository. One sec.

PR #3, opened 2015-10-09
AG cloning for test DB

squirrelo:
This pulls down and runs the `ag make test` from the american-gut-web repo so we have a full test database to use.


squirrelo:
That...wasn't ready for merge.

josenavas:
For future reference - use "WIP" to specify that a PR is not ready to merge - otherwise we assume that it is ready.

mortonjt:
:+1: Sorry about that

PR #6, opened 2015-10-10
Make test db

squirrelo:


squirrelo:
Now this is ready

mortonjt:
I think this looks good

.travis.yml
```diff
@@ -10,9 +10,17 @@ before_install:
...
   - travis_retry pip install .
 script:
+  - git clone https://github.com/biocore/american-gut-web.git ~/build/biocore/american-gut-web
+  - cd /home/travis/build/biocore/american-gut-web
+  - export AG_CONFIG=`pwd`/ag_config.txt.example
+  - export PYTHONPATH=/home/travis/build/biocore/american-gut-web:$PYTHONPATH
+  - ./scripts/ag make test
+  - export PYTHONPATH=
+  - $PYTHONPATH
+  - cd ~/build/biocore/labadmin/knimin/
```
wasade: There should be some magical travis env variable for this. See [here](http://docs.travis-ci.com/user/environment-variables/) for a list

knimin/config.txt.example
```diff
@@ -3,9 +3,9 @@ debug = True
 help_email =
 
 [postgres]
-USER =
+USER = postgres
 PASSWORD =
-DATABASE =
+DATABASE = ag_test
```
wasade: since labadmin is not ag, would it be possible to rename this? i think it would be a great policy going forward to avoid nomenclature that ties this to ag

wasade: ...except where necessary/appropriate

wasade:
2 minor comments, :+1: otherwise

wasade:
2 minor comments, :+1: otherwise

knimin/config.txt.example
squirrelo: The database made by running `ag_make_test` i called ag_test, so can't really rename.

wasade: Well, itd be annoying at least... Okay, not blocking

squirrelo:
Ok, that env variable is cool. Ready to go.

wasade:
:+1: for merge

PR #7, opened 2015-10-12
Add empty blanks list

squirrelo:
quick fix for the barcode scanner after the pulldown function changes.


mortonjt:
Is this good to go?

squirrelo:
yeah, was a tiny error

PR #8, opened 2015-10-13
Check if blanks exist

squirrelo:
This fixes the issue with an extra blank line being added to the pulldown template if no blanks are passed.


wasade:
Blanks are only relevant for the human survey?

squirrelo:
I don't have a way to figure out what blanks are on what plate right now, so this is the compromise I made.

wasade:
But why is the support for blanks tied to the human survey when blanks are
independent of sample type and associated host?

squirrelo:
Because we mostly deal in humans, so there will presumably be human barcodes pulled down every time.

rob-knight:
That doesn’t make sense to me; many different projects contain blanks, and many runs have no human samples.

squirrelo:
Ok, I’ll work out something. This may have to wait until we have the plate mapper in place. Also, right now thus pulldown is just AG, so we can make a few assumptions off that.

wasade:
Oh, I see, the issue is that the set of input sample IDs can span multiple
survey types so it isn't clear if the blank information should go with the
human, pet or environmental surveys. And, it isn't clear how to fit those
surveys into a common sample template.

I'm inclined to agree here in that we solve the immediate issue as I think
the larger problem will block the need this PR is fulfilling.

squirrelo:
That’s exactly it.

mortonjt:
Are these tests supposed to fail?

squirrelo:
yes

mortonjt:
I'm not feeling comfortable merging this, given that all of the other PRs tests are passing.  @ElDeveloper @josenavas would you like to double check this?

squirrelo:
eh, can do a pull and it should pass. This was done before tests got passing and just langushed.

squirrelo:
passing

PR #13, opened 2015-10-23
Informative pulldown failure

squirrelo:
Now you have reasons for why barcodes failed to pull down.


knimin/lib/data_access.py
```diff
@@ -827,6 +827,104 @@ def pulldown(self, barcodes, blanks):
...
+        def update_failure(sql, all_bc, reason):
+            hold = [x[0] for x in
+                    self._con.execute_fetchall(sql, [tuple(all_bc)])]
+            remaining_bc = all_bc.difference(hold)
+            return [(bc, reason) for bc in hold], remaining_bc
+
+        failure_info = []
+        current_barcodes = set(barcodes)
+
+        # TEST ORDER HERE MATTERS!
```
wasade: ?

knimin/handlers/ag_pulldown.py
```diff
@@ -41,8 +41,10 @@ def post(self):
         metadata, failures = db.pulldown(barcodes, blanks)
 
         meta_zip = InMemoryZip()
-        failtext = ("The following barcodes were not retrieved for any "
-                    "survey:\n%s" % "\n".join(failures))
+        failed = '\n'.join(['\t'.join(x) for x in
+                            db.explain_pulldown_failures(failures)])
```
wasade: why is a separate method call needed, can't this just be reported during the `pulldown` on like 41?

knimin/lib/data_access.py
```diff
@@ -827,6 +827,104 @@ def pulldown(self, barcodes, blanks):
...
+        # No more unexplained, so done
+        if len(current_barcodes) == 0:
+            return failure_info
+
+        # handout barcode
+        sql = """SELECT barcode
+                 FROM ag.ag_handout_barcodes
+                 WHERE barcode IN %s"""
+        failures, current_barcodes = update_failure(
+            sql, current_barcodes, 'Withdrawn sample')
```
wasade: how is this a withdrawn sample given that the query below is more explicit on being a withdrawn sample?

knimin/handlers/ag_pulldown.py
squirrelo: I suppose it could, but the pulldown code relies on the list returned by pulldown to check if the pulldown works for a specific barcode or not. I could check by length of failures rather than actual value.

knimin/lib/data_access.py
squirrelo: It's a cascade of tests, so removal order matters, e.g. first it removes all non-AG barcodes so you can assume the remaining ones belong to AG. Then it removes handout ones, so you can assume the rest are assigned to registered kits, etc.

knimin/lib/data_access.py
squirrelo: Copy-paste error. Fixed

knimin/lib/data_access.py
```diff
@@ -823,9 +823,112 @@ def pulldown(self, barcodes, blanks):
...
+        def update_failure(sql, all_bc, reason):
+            hold = [x[0] for x in
+                    self._con.execute_fetchall(sql, [tuple(all_bc)])]
+            remaining_bc = all_bc.difference(hold)
+            return [(bc, reason) for bc in hold], remaining_bc
+
+        failure_info = []
+        current_barcodes = set(barcodes)
+
+        # TEST ORDER HERE MATTERS! Assumptions made based on filtering of
```
wasade: Is it better to remove the barcodes and only flag a single error, or to flag all errors associated with a barcode? (Just asking, I do not know what is best here)

squirrelo: presumably they only fail because of a single reason, so filtering works here.

knimin/lib/data_access.py
```diff
@@ -823,9 +823,112 @@ def pulldown(self, barcodes, blanks):
...
+        # curent_barcodes by previous checks
+        # not an AG barcode
+        sql = """SELECT barcode
+                FROM ag.ag_kit_barcodes
```
mortonjt: Indent?

knimin/lib/data_access.py
```diff
@@ -823,9 +823,112 @@ def pulldown(self, barcodes, blanks):
...
+
+    def _explain_pulldown_failures(self, barcodes):
+        """Builds failure reason list for barcodes passed
+
+        Parameters
+        ----------
+        barcodes : list of str
+            Barcodes to explain failure for
+
+        Returns: list of tuples of str
```
mortonjt: Format to
```
Returns
-------
```
?

knimin/handlers/barcode_util.py
```diff
@@ -147,44 +147,33 @@ def get_ag_details(self, barcode):
...
-                               "never happen. Please notify "
-                               "someone on the database crew.")
-                    ag_details['email_type'] = "-1"
+
+            # it has all sample details
+            # (sample time, date, site)
+            if failures:
+                div_id = "no_metadata"
+                message = "Cannot retrieve metadata: %s" % failures[0][1]
+                ag_details['email_type'] = "-1"
```
wasade: why is this a magic string of a magic number?

knimin/handlers/barcode_util.py
```diff
@@ -147,44 +147,33 @@ def get_ag_details(self, barcode):
...
+            elif survey_type[survey_id] == 'Human':
```
wasade: what happens with environmental?

knimin/handlers/barcode_util.py
squirrelo: I have no idea, it was like that from the beginning and I just removed an if statement wrapping it.

knimin/handlers/barcode_util.py
squirrelo: caught by the `if failures:` since one of the reasons is "environmental sample"

knimin/lib/data_access.py
```diff
@@ -823,9 +823,114 @@ def pulldown(self, barcodes, blanks):
...
+        if len(barcodes) == 0:
+            return []
+
+        def update_failure(sql, all_bc, reason):
+            hold = [x[0] for x in
+                    self._con.execute_fetchall(sql, [tuple(all_bc)])]
+            remaining_bc = all_bc.difference(hold)
+            return [(bc, reason) for bc in hold], remaining_bc
+
+        failure_info = []
```
wasade: if this was a `dict`, then `update_failures` is not nearly as confusing as you'll be able to just do set operations against `failure_info`. basically, there wouldn't be a need to return `remaining_bc` and calls would then look like:

```python
failure_info.update(update_failures(sql, reason))
current_barcodes = current_barcodes - failure_info
```

knimin/lib/data_access.py
```diff
@@ -823,9 +823,114 @@ def pulldown(self, barcodes, blanks):
...
+
+        # environmental sample
+        sql = """SELECT barcode
+                 FROM ag.ag_kit_barcodes
+                 WHERE environment_sampled IS NOT NULL AND barcode in %s"""
+        failures, current_barcodes = update_failure(
+            sql, current_barcodes, 'Environmental sample')
+        failure_info.extend(failures)
+        # No more unexplained, so done
+        if len(current_barcodes) == 0:
```
wasade: no real benefit in the short circuits as there won't be callouts to the db if `current_barcodes` is empty. or, rather, i imagine that's how psycopg2 would handle passing the empty `list`/`set?

squirrelo: Unfortunately it actually does do the call and fails, since it passes `AND barcode in ()` as the SQL.

knimin/lib/data_access.py
squirrelo: Ah, thanks, better logic there. done.

squirrelo:
Comments addressed, should be good to go

PR #18, opened 2015-10-28
Use dupes lookup for host_subject_id

squirrelo:
Requires [american-gut-web #542](https://github.com/biocore/american-gut-web/pull/542)

Uses dupes lookup table to standardize participant names for host_subject_id creation

NOTE: this will fail until #542 is merged. Restart the tests to make it not fail.

### DIFF
--- a/knimin/lib/data_access.py
+++ b/knimin/lib/data_access.py
@@ -687,7 +687,9 @@ class KniminAccess(object):
             md[1][barcode]['HOST_COMMON_NAME'] = md_lookup[site]['COMMON_NAME']
 
             participant_name = dupes_lookup.get(
-                md[1][barcode]['SURVEY_ID'], specific_info['participant_name'])
+                md[1][barcode]['SURVEY_ID'],
+                specific_info['participant_name']).lower()
+
             md[1][barcode]['HOST_SUBJECT_ID'] = sha512(
                 specific_info['ag_login_id'] + participant_name).hexdigest()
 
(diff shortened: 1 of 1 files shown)


mortonjt:
Would it be worthwhile to add a test case to the data_access tests?

squirrelo:
we don’t really have data access tests for this currently. Also, that would requirere factoring and tests for the entirety of the pulldown code which, while planned, is a logistical nightmare outside the scope of this pul request.

mortonjt:
I'm not comfortable merging this without tests.  Any thoughts @wasade, @ElDeveloper @josenavas ?

knimin/lib/data_access.py
```diff
@@ -679,9 +685,11 @@ def format_survey_data(self, md):  # noqa
             md[1][barcode]['BODY_PRODUCT'] = md_lookup[site]['BODY_PRODUCT']
             md[1][barcode]['DESCRIPTION'] = md_lookup[site]['DESCRIPTION']
             md[1][barcode]['HOST_COMMON_NAME'] = md_lookup[site]['COMMON_NAME']
+
+            participant_name = dupes_lookup.get(
+                md[1][barcode]['SURVEY_ID'], specific_info['participant_name'])
             md[1][barcode]['HOST_SUBJECT_ID'] = sha512(
-                specific_info['ag_login_id'] +
-                specific_info['participant_name']).hexdigest()
+                specific_info['ag_login_id'] + participant_name).hexdigest()
```
wasade: is the name, email, etc lowered?

squirrelo: Nope, name was not lowercased. Fixed.

squirrelo: This shouldn't affect things though, as the case-sensitive names are also in the duplicates table.

squirrelo:
Comment addressed

wasade:
@mortonjt, more formal testing is being added in but as @squirrelo points out, there is quite a bit of logistics to go through and realistically will take a few weeks. Unit testing is not the only way to test code, but it is of course a powerful way.

PR #19, opened 2015-10-28
mapping file fixes

squirrelo:
This is a collection of all fixes needed before updates to EBI

### DIFF
--- a/knimin/lib/util.py
+++ b/knimin/lib/util.py
@@ -198,6 +198,26 @@ def categorize_age(x):  # noqa
     return age_cat
 
 
+def correct_age(age, height, weight, etoh):
+    """Infers incorrect ages and incorrectly classified babies"""
+    # Checks the logic for age
+    if age >= 4 and age < 123:
+        return age
+    if age < 0 or age >= 123:
+        return 'Unspecified'
+
+    # Checks the logic for height (in cm)
+    if height != 'Unspecified' and height > 91.4:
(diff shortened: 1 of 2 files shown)


mortonjt:
Are the tests supposed to fail?

knimin/lib/util.py
```diff
@@ -198,6 +198,26 @@ def categorize_age(x):  # noqa
     return age_cat
 
 
+def correct_age(age, height, weight, etoh):
+    """Infers incorrect ages and incorrectly classified babies"""
+    # Checks the logic for age
+    if age >= 4 and age < 123:
```
mortonjt: Why can't we consider ages between 0 and 4?

squirrelo: those are the ages we are making sure are correct through the height, weight, and alcohol checks.

mortonjt: Alright! Makes sense.

Could you rename the function to something like `fix_babies_age` or something?

mortonjt:
One minor comment, then I think it is ready to merge.

knimin/lib/util.py
squirrelo: it's not just babies though, it's also highlanders (people over 123 years old) and time travelers (people with negative ages)

mortonjt: ahhh :L 

ok cool.  I think this is good

PR #22, opened 2015-10-30
Unicode transform for answers

squirrelo:
Quick fix for new unicode issues on pulldown.


squirrelo:
This is currently testing with a DB pulldown from a week ago and so far so good.

PR #23, opened 2015-10-30
More pulldown fixes

squirrelo:
- Fixes issue with unicode conversion and stripping characters
- Uppercases the zipcode so foreign zipcodes are not case sensitive
- Converts collection month to string for consistency with old metadata format
- Unreported weight, height, and BMI from these all now set to 'Unspecified'
- Make definition of baby age range consistent

### DIFF
--- a/knimin/lib/data_access.py
+++ b/knimin/lib/data_access.py
@@ -14,8 +14,8 @@ from psycopg2.extras import DictCursor
 
 from util import (make_valid_kit_ids, make_verification_code, make_passwd,
                   categorize_age, categorize_etoh, categorize_bmi, correct_age)
-from constants import (md_lookup, month_lookup, season_lookup,
-                       regions_by_state, blanks_values)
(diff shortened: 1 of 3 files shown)


squirrelo:
This is ready for review

knimin/lib/data_access.py
```diff
@@ -612,7 +625,7 @@ def format_survey_data(self, md):  # noqa
             md[1][barcode]['REQUIRED_SAMPLE_INFO_STATUS'] = 'completed'
 
             # Sample-dependent information
-            zipcode = md[1][barcode]['ZIP_CODE']
+            zipcode = md[1][barcode]['ZIP_CODE'].upper()
```
mortonjt: I'm curious, what sort of zip codes does this apply to?

squirrelo: internationals, for example `WC2E 9RZ` is in london.

mortonjt:
:+1:

PR #24, opened 2015-11-02
Vioscreen pulldown

squirrelo:
Allows third party surveys to be pulled down as part of the full metadata.


knimin/lib/data_access.py
```diff
@@ -504,6 +507,8 @@ def format_survey_data(self, md):  # noqa
         -------
         dict of dict of dict
```
mortonjt: Not necessary for this PR (since this was here before)

But we really need to specify what exactly the keys/values are.

squirrelo: Agree, this will probably be 5-10 ancillary functions in the end, so will be good to know what each dict holds.

squirrelo:
Test added and passing, ready for merge.

PR #29, opened 2015-11-03
Make blanks all false for subset categories

squirrelo:


squirrelo:
This just needs a quick once over and merge.

PR #32, opened 2015-11-07
More col removal

squirrelo:
- Removes more columns for EBI release
- Allows full pulldown of all cols if wanted
- Fixes host_common_name

### DIFF
--- a/knimin/lib/data_access.py
+++ b/knimin/lib/data_access.py
@@ -17,7 +17,8 @@ from psycopg2.extras import DictCursor
 from util import (make_valid_kit_ids, make_verification_code, make_passwd,
                   categorize_age, categorize_etoh, categorize_bmi, correct_age)
 from constants import (md_lookup, month_int_lookup, month_str_lookup,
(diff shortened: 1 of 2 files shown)


knimin/lib/data_access.py
```diff
@@ -489,7 +490,7 @@ def _months_between_dates(self, d1, d2):
         # Calculate the number of 12-month periods between the years
         return (d2.year - d1.year) * 12 + (d2.month - d1.month)
 
-    def format_survey_data(self, md, external_surveys=None):  # noqa
+    def format_survey_data(self, md, external_surveys=None, full=False):  # noqa
```
mortonjt: Would be nice to update the docs

mortonjt:
One minor comment.  Otherwise, its ready to merge

knimin/lib/data_access.py
squirrelo: Again, this is code that will be refactored as soon as it works and we can add tests, so documentation at this point is moot.

squirrelo:
Ready for merge

PR #34, opened 2015-11-30
More metadata cleaning for EBI

squirrelo:
Removes the following columns:
'DIET_RESTRICTIONS_LIST'
'OTHER_CONDITIONS_LIST'
'RACE_OTHER'


EmbrietteH:
Looks ok, but are we 100% sure now that we have all free text columns into which someone could write a person's name removed with this PR? This had been checked previously by several and these three columns were missed, so I want to make sure we don't make the same mistake again. Thanks!

squirrelo:
All freetext fields are as follows: 
 PREGNANT_DUE_DATE
 MEDICATION_LIST
 DOG_LOCATION
 RACE_OTHER
 SUPPLEMENTS
 DOG_CONTACT
 HEIGHT_CM
 WEIGHT_KG
 ZIP_CODE
 ABOUT_YOURSELF_TEXT
 CAT_LOCATION
 DIET_RESTRICTIONS_LIST
 TRAVEL_LOCATIONS_LIST
 RELATIONSHIPS_WITH_OTHERS_IN_STUDY
 CAT_CONTACT
 ANTIBIOTIC_MED
 ANTIBIOTIC_CONDITION
 NAME
 ANIMAL_AGE
 ANIMAL_FREE_TEXT
 ANIMAL_TYPE_FREE_TEXT
 OTHER_ANIMALS_FREE_TEXT
 HUMANS_FREE_TEXT
 OTHER_CONDITIONS_LIST

squirrelo:
Ok, everything freetext save height and weight is removed, along with some other non-freetext identifying fields.

squirrelo:
No other comments so taking this off WIP

EmbrietteH:
I see 'gender' in the code but that is not free text so should not be included in the remove list, correct?
@mortonjt can you please also review this? We need this merged and deployed for round 22 which came off the sequencer today.

squirrelo:
According to @ackermag we need to replace gender with sex as the column header, so this removes the gender column and earlier code copies it over to sex.

EmbrietteH:
From the list that you commented above, the following are not in the code:

'ANIMAL_AGE',
 'ANIMAL_FREE_TEXT',
 'ANIMAL_TYPE_FREE_TEXT',
 'HUMANS_FREE_TEXT',
 'NAME',
 'OTHER_ANIMALS_FREE_TEXT',

I assume that 'NAME' is removed in another location in the code, but I want to ensure that these others have also been removed. If so, good to merge.

EmbrietteH:
Just had Jose check, confirmed those columns mentioned in my previous comment are not pulling down; this PR is good to merge.

squirrelo:
Name is part of the animal survey, so we don't need to worry about it here. Deployed!

PR #35, opened 2015-12-09
WIP - Add plate mapper functionality

squirrelo:
This branch is the plate mapper and tracking system. The database changes are currently in another branch in the american-gut-web repository as well. 

The idea of the system is that we have protocols that can be broken down into discrete steps. Each step has one or more instrument typess, instrument peripheral types (e.g. the pipette head used on a robot), reagents, and primers attached to the step. These steps are then performed on plates or groups of samples, and the actual performing of the protocol stores the lot number, instrument serial, etc. of the items used to perform the steps. These protocol performances are then pooled together in the pool object, and attached to a run object for sequencing. This way we can track everything from sample initial scan in to the final sequencing run and prep template generation. 

The protocol, step, *Type and accompanying specific classes are read only. That is, once an item is added to the system, the only way to edit it is to delete and create anew one. This allows us to have a non-changing tracking of what is going on, e.g. you say protocol 1 is used, and protocol 1 can never be edited so we always know what you did. 

Performance is editable as you add protocols and samples to it, and pools/runs are editable as you add performances and settings to them. Sample and plate objects are editable as you add metadata and collect sampels onto them. All these will have a `finalize` call that then locks them into a read-only mode so they can't be changed from that point out. Again, this is so we can maintain data integrity after a plate is full and run on a protocol, etc.
TODO:
- [ ] Protocol, Performance, and Step objects
- [ ] InstrumentType and Instrument objects
- [ ] PeripheralType and Peripheral objects
- [ ] ReagentType and Reagent objects
- [ ] PrimerType and Primer objects
- [ ] Person and User objects
- [ ] Sample and Plate objects
- [ ] Pool and Run objects
- [ ] Interface (To be broken down later)

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,3 +1,3 @@
 Knight Lab administrative portal
 
-This is the internal Knight Lab administration portal for creating, tracking, and processing general barcodes and the American Gut.
+This is the internal Knight Lab administration portal for creating, tracking, and processing general barcodes and the American Gut, along with full source tracking from sample arrival through sequencing.


squirrelo:
Closed because we are doing seperate plate mapper for this

PR #36, opened 2016-01-04
Add ability to change year and month for postmark date

squirrelo:
Was brought up that the datepicker for postmark date does not allow you to go back to a different year. This fixes that, so we can log samples across new years changes.

### DIFF
--- a/knimin/lib/tests/test_geocoder.py
+++ b/knimin/lib/tests/test_geocoder.py
@@ -70,8 +70,8 @@ class TestCallWrapper(TestCase):
 class TestGeocode(TestCase):
     def test_geocode_nonmock(self):
         obs = geocode('9500 Gilman Dr, La Jolla, CA')
-        exp = Location('9500 Gilman Dr, La Jolla, CA', 32.8794081,
-                       -117.2368167, 115.2523956298828, 'San Diego', 'CA',
+        exp = Location('9500 Gilman Dr, La Jolla, CA', 32.8794239,
+                       -117.2369135, 114.7895736694336, 'San Diego', 'CA',
                        '92093', 'United States')
         self.assertEqual(obs, exp)
 
--- a/knimin/templates/barcode_util.html
+++ b/knimin/templates/barcode_util.html
@@ -7,7 +7,9 @@
     $(".chosen-select").chosen();
     $(".date-picker").datepicker({
       changeMonth: true,
-      maxDate: '+0m',
+      changeYear:true,
+      minDate: '-4m',
+      maxDate: '-0',
       onSelect: function(dateText, inst) {
           $(this).focus();
       }


squirrelo:
Ping @mortonjt

squirrelo:
Google updated their lat/longs, tests updated for this and now passing.

mortonjt:
I think it is ok.

Now, is this using the Google Maps API?  We might need to be careful, because that API has recently become deprecated.

squirrelo:
It has been for a long time now.

mortonjt:
Yeah.  Google maps is amazing.  But it looks like Google maps will be no longer supported as of this January :(

https://developers.google.com/maps-engine/new
http://googlegeodevelopers.blogspot.com/2014/12/announcing-deprecation-of-google-earth.html

rob-knight:
What?! Are they replacing it with something else?

wasade:
In the immortal words of Cartman: lame

antgonza:
From: http://blog.mangomap.com/post/108790608911/the-end-of-google-maps-engine-causes-mass

What’s the best alternative? Mango of Course!

In the email announcement that Google sent to all account holders they encouraged users to either move to the very lightweight Google MyMaps or use their existing range of API products. But where does that leave non-coders and users who need more than the simple pushpin on a map functionality that MyMaps offers?

The short answer is that you need a non-Google alternative and luckily for you at Mango we have everything you need and are looking forward to welcoming you to our family of over 15,000 map makers!

We offer all of the functionality available on Google Maps Engine and much more.

rob-knight:
So do we use the google map engine or the google map API (which is not
affected)?

On Jan 6, 2016, at 7:44 AM, Antonio Gonzalez <notifications@github.com>
wrote:

From:
http://blog.mangomap.com/post/108790608911/the-end-of-google-maps-engine-causes-mass

What’s the best alternative? Mango of Course!

In the email announcement that Google sent to all account holders they
encouraged users to either move to the very lightweight Google MyMaps or
use their existing range of API products. But where does that leave
non-coders and users who need more than the simple pushpin on a map
functionality that MyMaps offers?

The short answer is that you need a non-Google alternative and luckily for
you at Mango we have everything you need and are looking forward to
welcoming you to our family of over 15,000 map makers!

We offer all of the functionality available on Google Maps Engine and much
more.

—
Reply to this email directly or view it on GitHub
<https://github.com/biocore/labadmin/pull/36#issuecomment-169362882>.

mortonjt:
According to @squirrelo , we are using the Google Maps API.  So I think we are in the clear.  Sorry for the misunderstanding.

squirrelo:
Correct, google maps API will still be up. It's google Earth going away. In light of that, this is ready to merge.

PR #37, opened 2016-01-19
Remove willing to be contacted col

squirrelo:
This edits the ebi removal columns and also adds a try/except so if we try and delete a retired question for EBI pulldown, the system does not error.

### DIFF
--- a/knimin/lib/data_access.py
+++ b/knimin/lib/data_access.py
@@ -793,8 +793,9 @@ class KniminAccess(object):
                 for col in ebi_remove:
                     try:
                         del md[1][barcode][col]
-                    except:
-                        # Column doesn't exist already, so no removal needed
+                    except KeyError:
+                        # Column doesn't exist already for survey (retired),
+                        # so no removal needed
                         pass
 
             # Add the external surveys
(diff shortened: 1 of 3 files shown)


knimin/lib/data_access.py
```diff
@@ -791,7 +791,11 @@ def format_survey_data(self, md, external_surveys=None, full=False):  # noqa
             # Get rid of columns not wanted for pulldown
             if not full:
                 for col in ebi_remove:
-                    del md[1][barcode][col]
+                    try:
+                        del md[1][barcode][col]
+                    except:
+                        # Column doesn't exist already, so no removal needed
```
mortonjt: This is just for debugging purposes in the future.

What do you think about adding a warning, indicating that a column that doesn't exist was attempted to be deleted?

mortonjt:
:+1: with one minor comment

knimin/lib/data_access.py
squirrelo: Probably not useful, since the way this happens is a question is retired, and therefore not in the pulldown any more, but the EBI removal list is for all iterations of the survey. All it would tell you is the person has a newer version of the survey.

PR #38, opened 2016-01-21
Consent checker

squirrelo:
This adds a very bare-bones page for checking if one or more barcodes have consent attached. It also gives granular reasons for why a barcode does not have consent, same as the barcode util page.


squirrelo:
This still needs some cosmetic work but the functionality should all be there.

squirrelo:
Ok, this is good to go

mortonjt:
:+1:

PR #40, opened 2016-01-27
Explicit conversions to float

squirrelo:
Some items were being passed into the conversion functions as strings, which messed up the if statements. This makes the floats explicit in the code.

### DIFF
--- a/knimin/lib/util.py
+++ b/knimin/lib/util.py
@@ -172,25 +172,30 @@ def make_verification_code(vercode_length=5):
 
 
 def categorize_age(x):  # noqa
-    if x < 0:
+    if x == 'Unspecified':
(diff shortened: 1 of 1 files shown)


knimin/lib/util.py
```diff
@@ -232,16 +246,21 @@ def categorize_etoh(x):
 
 
 def categorize_bmi(x):
-    if x < 0:
+    if x == 'Unspecified':
+        return 'Unspecified'
```
wasade: if this is all about casting to float, shouldn't this be `nan`?

squirrelo: The issue is it comes from the conversions, so it can be a float or 'Unspecified' if it can't be calculated, thus this test is needed.

PR #42, opened 2016-02-04
Add corrected BMI column to pulldown

squirrelo:
Does what it says on the tin. Also changes the sane BMI range to 8-79.


josenavas:
Merge conflict

PR #44, opened 2016-02-15
Zip in full pulldown

squirrelo:
Adds full pulldown capability to the command line interface.

### DIFF
--- a/knimin/lib/data_access.py
+++ b/knimin/lib/data_access.py
@@ -520,11 +520,18 @@ class KniminAccess(object):
         barcode_info = self.get_ag_barcode_details(all_barcodes)
 
         # tuples are latitude, longitude, elevation, state
-        zipcode_sql = """SELECT UPPER(zipcode), country,
-                             round(latitude::numeric, 1),
-                             round(longitude::numeric,1),
-                             round(elevation::numeric, 1), state
-                         FROM zipcodes"""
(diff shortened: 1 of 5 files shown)


knimin/lib/data_access.py
```diff
@@ -675,15 +682,22 @@ def format_survey_data(self, md, external_surveys=None, full=False):  # noqa
...
                 if info.lat is not None:
-                    md[1][barcode]['LATITUDE'] = "%.1f" % info.lat
-                    md[1][barcode]['LONGITUDE'] = "%.1f" % info.long
-                    md[1][barcode]['ELEVATION'] = "%.1f" % info.elev
+                    if full:
+                        md[1][barcode]['LATITUDE'] = info.lat
+                        md[1][barcode]['LONGITUDE'] = info.long
+                        md[1][barcode]['ELEVATION'] = info.elev
+                    else:
+                        md[1][barcode]['LATITUDE'] = round(info.lat, 1)
```
wasade: why round twice?

squirrelo: This  rounds the newly geocoded data, while the SQL above gets old geocode data that was stored in the DB.

mortonjt:
Normally, I'd say that the `pulldown` method needs a test case.  

Otherwise :+1:

PR #45, opened 2016-02-18
Mark ebi submitted

squirrelo:
This adds a button to click that updates the barcodes deposited status by pulling down all barcodes in EBI and updating from that list. This is what is needed for PR [#571](https://github.com/biocore/american-gut-web/pull/571) in american-gut-web, so that can also be reviewed and merged at this point.

### DIFF
--- a/knimin/lib/data_access.py
+++ b/knimin/lib/data_access.py
@@ -2027,12 +2027,13 @@ class KniminAccess(object):
             'http://www.ebi.ac.uk/ena/data/warehouse/filereport?accession='
             '%s&result=read_run&fields=sample_alias' % accession)
         # Clean EBI formatted sample names to just the barcodes
-        barcodes = [s.strip().split(':')[1].split('.')[1] for s in samples]
+        barcodes = tuple(s.strip().split(':')[1].split('.')[1]
+                         for s in samples)
 
         sql = """UPDATE ag.ag_kit_barcodes
                  SET deposited = TRUE
                  WHERE barcode IN %s"""
-        self._con.execute(sql, [tuple(barcodes)])
+        self._con.execute(sql, [barcodes])
 
     def _clear_table(self, table, schema):
         """Test helper to wipe out a database table"""
--- a/knimin/lib/data_access.py
+++ b/knimin/lib/data_access.py
@@ -1892,7 +1892,8 @@ class KniminAccess(object):
                     cast(ag_kit_id as varchar(100)), barcode,  site_sampled,
                     environment_sampled, sample_date, sample_time,
                     participant_name, notes, refunded, withdrawn, moldy, other,
-                    other_text, date_of_last_email ,overloaded, name, status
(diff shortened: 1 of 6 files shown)


wasade:
Biggest hesitation is this could easily get out of sync with qiita's
perspective, although I suspect that a remote API for checking these things
with qiita is a bit out...

josenavas:
@wasade in which sense this goes out of sync from qiita? Meaning that Qiita will have more data than labadmin? (I can't see how the other way around can happen).

Should be easily fixed by a cron job running once a day.

wasade:
out of sync in the sense that human error is a high possibility

squirrelo:
What human error? This reads directly from EBI, so there is no human involvement in the barcode information.

wasade:
Ah, I see that now! Sorry for the confusion

josenavas:
I did not had time to go over the code, but I want to double check that this case have been covered. Does this code works with those barcodes that a letter has been added at the end (e.g. A, B, a, b)?

There are a few cases of those in the early rounds of AG and I think they appear in EBI with such suffix.

squirrelo:
Ah, good catch, just pushed a fix for that.

knimin/lib/data_access.py
```diff
@@ -2018,6 +2020,22 @@ def getProjectNames(self):
...
+    def set_deposited_ebi(self):
+        """Updates barcode deposited status by checking EBI"""
+        accession = 'ERP012803'
+        samples = fetch_url(
+            'http://www.ebi.ac.uk/ena/data/warehouse/filereport?accession='
+            '%s&result=read_run&fields=sample_alias' % accession)
+        # Clean EBI formatted sample names to just the barcodes
+        # stripped of any appended letters for barcodes run multiple times
+        barcodes = tuple(s.strip().split('.')[1][:9]
+                         for s in samples if len(s.split('.')) == 2)
```
wasade: Wonder if a regex would be more explicit here? Example output the resource request is below. I think a sufficient regex would be:

```python
13:20:15 (dtmcdonald@barnacle):ag_qiita_seqs_15feb2015> s = 'qiita_sid_10317:10317.000001031'

13:20:25 (dtmcdonald@barnacle):ag_qiita_seqs_15feb2015> re.compile('(\d){9}$').search(s).group(0)
Out[16]: '000001031'
```

Example text from EBI:

```text
sample_alias
qiita_sid_10317:10317.000001002
qiita_sid_10317:10317.000001004
qiita_sid_10317:10317.000001008
qiita_sid_10317:10317.000001018
qiita_sid_10317:10317.000001028
qiita_sid_10317:10317.000001031
qiita_sid_10317:10317.000001032
qiita_sid_10317:10317.000001038
```

squirrelo: I just tested this for adding an A or B to the end of some of those and it doesn't work. I've tried a few other regex combinations but none seem to pick up both the barcode alone and the letter appended barcode.

PR #54, opened 2016-04-20
Give benefit of the doubt if missing info

squirrelo:
This modifies the logic for correct age to not assume undefined age or weight means we can't correct the age.


wasade:
what motivated this?

squirrelo:
Talk with @JWDebelius to get it in line with original notebook.

EmbrietteH:
@JWDebelius please check and verify this is in line with what you need.

jwdebelius:
Let's hold off until the next meeting, as we sort out how we're going to define subsets. I think the age_corrected probably also falls into a subset, and may be better defined there...

squirrelo:
Closed in favor of moving this out to a notebook.

PR #55, opened 2016-04-25
Add access controls

squirrelo:
This breaks apart functionality in labadmin into discrete access groups that must be set per user.

Requires [american gut #587](https://github.com/biocore/american-gut-web/pull/587)

### DIFF
--- a/knimin/handlers/ag_add_barcode_kit.py
+++ b/knimin/handlers/ag_add_barcode_kit.py
@@ -5,7 +5,7 @@ from knimin.handlers.access_decorators import set_access
 from knimin import db
 
 
-@set_access(['Create AG kits'])
+@set_access(['AG kits'])
 class AGAddBarcodeKitHandler(BaseHandler):
     @authenticated
     def get(self):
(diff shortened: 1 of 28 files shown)


knimin/handlers/ag_add_barcode_kit.py
```diff
@@ -7,11 +7,13 @@
 class AGAddBarcodeKitHandler(BaseHandler):
```
wasade: would it be possible to have the methods default to a high access level? In other words, unless explicitly stated, you need `Admin`. It's conservative in the event some handlers get over looked, or future development forgets to add the `self.has_access` calls

squirrelo: I can add that, although that requires adding a base access level to the database as well.

wasade: So then general forms would have something like `self.has_access('basic')`?

squirrelo: yup

squirrelo:
This is ready for review, although I can't add tests until the AG DB PR is in. Also, acronyms.

squirrelo:
ping @EmbrietteH @wasade

wasade:
seems good

PR #56, opened 2016-04-27
Remove errant handout kits

squirrelo:
This adds a listing of all handout kits in the system, using DataTables, that can be exported to excel format or used to delete errant kits. 

Requires #55

### DIFF
--- a/knimin/lib/data_access.py
+++ b/knimin/lib/data_access.py
@@ -1342,6 +1342,27 @@ class KniminAccess(object):
 
         return kits
 
+    def delete_ag_kits(self, email, kit_ids):
+        """Remove handout kits from the server
+
+        Parameters
+        ----------
+        email : str
+            Email of the user requesting the delete. Not strictly needed, but
+            asked for as a safety measure
+        kit_ids : list of str
+            Kits to delete
+        """
+        if not self.has_access(email, ['Admin', 'Create AG kits']):
+            raise ValueError('User %s does not have delete permissions!' %
+                             email)
+
+        bc_sql = "DELETE FROM ag.ag_handout_barcodes WHERE kit_id in %s"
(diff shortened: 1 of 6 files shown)


knimin/lib/data_access.py
```diff
@@ -2032,6 +2053,32 @@ def search_handout_kits(self, term):
...
+            Each kit as a dict with all available info. Barcodes are a list
+            under the barcode key.
+        """
+        sql = """SELECT kit_id, verification_code, swabs_per_kit, created_on,
+                 array_agg(barcode) as barcodes
+                 FROM ag.ag_handout_kits
+                 JOIN ag.ag_handout_barcodes USING (kit_id)"""
+        sql_args = None
+        if kit_ids is not None:
+            sql += 'WHERE kit_ids IN %s'
```
josenavas: Missing a space before `WHERE`. Probably better to create `sql` as a list and then do a `' '.join(sql)` at the end. I think list appending is cheaper than string concatenation.

josenavas:
I'm gonna be super-paranoic here. Will it make sense to write the deleted information into a log file or something so we can recover in case of undesired deletion? I may be missing something, but the delete shouldn't be checking that the kits have been already verified?

squirrelo:
This only deletes handout kits, so there's no verification needed that is programatic, only human to make sure they haven't been shipped yet. I agree on a log of some sort, but unsure how. Maybe just a TSV we keep appending to similar to other logging software, with the info for the given kit(s)?

josenavas:
Ok - yes the TSV that we append should be good enough - just add also the timestamp so it is easily searchable. I would also suggest to put some kind of size limitation (i.e. if the file is X bytes, close, add an incremental suffix, and open a new one).

PR #57, opened 2016-04-27
Stop reloading creating more kits. Fixes #30

### LINKED ISSUE #30: Refreshing "Add new barcodes" bug
After adding a set of barcodes to a new or existing project using the "Add new barcode" menu, an additional set of barcodes will be added without additional input if the page is refreshed.

squirrelo:
Stops accidental reloads of the create kits page from creating more kits. fixes #30.

### DIFF
--- a/knimin/handlers/ag_new_kit.py
+++ b/knimin/handlers/ag_new_kit.py
@@ -1,6 +1,6 @@
 #!/usr/bin/env python
-from json import dumps, loads
-from tornado.web import authenticated
+from json import loads
+from tornado.web import authenticated, HTTPError
 from knimin.handlers.base import BaseHandler
 from knimin import db
 from knimin.lib.mem_zip import InMemoryZip
@@ -52,8 +52,6 @@ class AGNewKitHandler(BaseHandler):
             kits = db.create_ag_kits(zip(num_swabs, num_kits), tag, projects)
             fields = ','.join(kits[0]._fields)
         except Exception as e:
-            msg = "ERROR: %s" % str(e)
-        else:
-            msg = "Kits created! Please wait for downloads."
+            raise HTTPError(500, "ERROR: %s" % str(e))
 
-        self.write({'kitinfo': dumps(kits), 'fields': fields, 'msg': msg})
+        self.write({'kitinfo': kits, 'fields': fields})
--- a/knimin/handlers/ag_new_kit.py
+++ b/knimin/handlers/ag_new_kit.py
@@ -56,9 +56,4 @@ class AGNewKitHandler(BaseHandler):
         else:
             msg = "Kits created! Please wait for downloads."
 
-        project_names = db.getProjectNames()
-        remaining = len(db.get_unassigned_barcodes())
-        self.render("ag_new_kit.html", projects=project_names,
-                    currentuser=self.current_user, msg=msg,
-                    kitinfo=dumps(kits), remaining=remaining,
-                    fields=fields)
+        self.write({'kitinfo': dumps(kits), 'fields': fields, 'msg': msg})
--- a/knimin/templates/ag_new_kit.html
+++ b/knimin/templates/ag_new_kit.html
@@ -36,25 +36,29 @@
     }
 
     function build_submit() {
+        $("#msg").text("")
         $("#submit-create").prop("disabled", true);
         if (!validate()) {
-            return false;
             $("#submit-create").prop("disabled", false);
+            return false;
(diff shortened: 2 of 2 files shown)


josenavas:
I was chatting with @elainewolfe today about this - isn't it easier and requires less changes to just redirect to the search page (that is what she does anyways to avoid double-creating kits) once the kits are created?

Also, I was concerned about the behavior of this - if it is a form post, most of the browsers asks the user if he is sure that he wants to resubmit the form. Any clue on why that was not happening here?

I think another concern too is that the kits gets created in the database, but the creation of the file takes a while. If for whatever reason the client submits the form, but gets disconnected from the server before the file gets created, is there a way to download the previously created file?

squirrelo:
The first two points do not involve what this PR is fixing. Originally, kit creation was done as a form submit so if the page reloaded the form submit happened again, meaning the kits were double created. Now the page submits with AJAX, so if the reload happens it does not create more kits. The resubmit was because the page reloaded.

Right now there is no way to recreate the kit printouts because the kit passwords are stored in bcrypted form. Once the changeover to user based logins happens these can be removed from being encrypted, but for now they are the same passwords users must use to login once the kit is registered, so it needs to be encrypted at all times.

antgonza:
I think @josenavas is right, wouldn't be easier to add a confirm('Are you sure you want to submit?') in the function validate and once the form is submitted move to search page so the user knows that the action is happening?

squirrelo:
I can add the confirm if it's really wanted, but the redirect will break the download of the files, and therefore cannot be done quickly or cleanly.

antgonza:
Thanks for confirming @squirrelo but I thought the download happened in a hidden iframe, is this not correct?

squirrelo:
It does, through the help of javascript and information passed from kit creation to that iframe, making the redirect extremely tricky if not impossible.

josenavas:
Can the download of the files be handled in the same way as it is done in Qiita?

squirrelo:
The file is made on the fly, so could you explain what you mean?

antgonza:
Got it, thanks for the explanations. I still think that the confirm/iframe is the best solution even if "complex". However, I can see how either solution is useless as the download creation job is blocking, right? With this in mind, perhaps it would be easier to just add the confirm to avoid resubmissions. Now, perhaps there is something I'm missing but not sure how the AJAX implementation is much better than a simple confirm?

squirrelo:
Again, this fixes the issue where reloading the page resubmits an already submitted form. A confirmation will not fix this, because you are reloading the POST, not the GET for the page.  By moving it to AJAX, that post only happens within the javascript of the page, so reloading does not cause the POST to fire again.

josenavas:
@squirrelo one of the concerns that we have, is that in Qiita we also have a lot of POSTs, but this issue does not appear there - so my guess is that there is something else wrong and it is worth investigating, as that can be happening in other parts of the system that we have not yet detected. Does it make sense?

squirrelo:
The difference here is that qiita posts do a discrete action, then move on. Here, the post creates the kit, then hands off to the download page that creates the files. This handoff is what is causing the issues, as the post will not ask for confirmation on reload because of this. The two are different scenarios.

squirrelo:
ping @EmbrietteH

EmbrietteH:
I think @josenavas @antgonza should take one more look...

antgonza:
Not sure what's new, there hasn't been any updates since our last review. Anyway, are we gonna add the submit confirmation button? BTW we could make the generation of codes not blocking by making them run in the back and simply have a list of barcodes created and it's timestamp of creation. This should allow for non blocking jobs, (*) have a way to keep track of sets of barcodes created, and later we could add more metadata. 

(*) I'm guessing currently there is no easy way to download the barcodes that were created, let's say, two weeks ago, right?

squirrelo:
There already is a confirmation button if you make a kit with more than 7 swabs per kit.

None of what has been proposed is within the scope of this PR. This was meant as a quick PR that addressed the single issue of reloading a page creating new kits, which is a bug that has caused a few isues already and needs to be fixed quickly. Other features can  be added in new PRs.

EmbrietteH:
Agree with @antgonza and @josenavas Let's add the confirm button and keep the other items for future PRs.

squirrelo:
Confirm updated to be universal now instead of greater than 7 swabs per kit.

squirrelo:
ping @EmbrietteH

josenavas:
👍

PR #58, opened 2016-05-02
Make updating barcodes work again

squirrelo:
This updates the code to allow changing site sampled, participant survey, etc for a barcode.

### DIFF
--- a/knimin/handlers/ag_edit_barcode.py
+++ b/knimin/handlers/ag_edit_barcode.py
@@ -14,7 +14,8 @@ class AGEditBarcodeHandler(BaseHandler):
             ag_login_id = db.search_kits(details['ag_kit_id'])[0]
             site_sampled = db.human_sites
             environment_sampled = db.general_sites
-            participants = db.getHumanParticipants(ag_login_id)
+            participants = db.getHumanParticipants(ag_login_id) + \
+                db.getAnimalParticipants(ag_login_id)
             self.render("ag_edit_barcode.html", response=None, barcode=barcode,
                         sites_sampled=site_sampled, details=details,
                         environments_sampled=environment_sampled,
@@ -42,10 +43,9 @@ class AGEditBarcodeHandler(BaseHandler):
             self.render("ag_edit_barcode.html", response='Good', barcode=None,
                         sites_sampled=None, details=None,
                         environments_sampled=None,
-                        logins=None, currentuser=self.current_user)
-        except Exception as e:
-            print str(e)
+                        participants=[], currentuser=self.current_user)
+        except:
             self.render("ag_edit_barcode.html", response='Bad', barcode=None,
                         sites_sampled=None, details=None,
                         environments_sampled=None,
-                        logins=None, currentuser=self.current_user)
+                        participants=[], currentuser=self.current_user)
--- a/knimin/handlers/ag_edit_barcode.py
+++ b/knimin/handlers/ag_edit_barcode.py
@@ -11,13 +11,15 @@ class AGEditBarcodeHandler(BaseHandler):
         barcode = self.get_argument('barcode', None)
         if barcode is not None:
             details = db.getAGBarcodeDetails(barcode)
+            ag_login_id = db.search_kits(details['ag_kit_id'])[0]
             site_sampled = db.human_sites
             environment_sampled = db.general_sites
-            logins = db.getAGKitsByLogin()
+            participants = db.getHumanParticipants(ag_login_id)
             self.render("ag_edit_barcode.html", response=None, barcode=barcode,
                         sites_sampled=site_sampled, details=details,
                         environments_sampled=environment_sampled,
-                        logins=logins, currentuser=self.current_user)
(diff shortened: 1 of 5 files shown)


EmbrietteH:
Can we get a quick review on this? It would be much appreciated! @antgonza @wasade @ElDeveloper

knimin/handlers/ag_edit_barcode.py
```diff
@@ -11,13 +11,15 @@ def get(self):
         barcode = self.get_argument('barcode', None)
         if barcode is not None:
             details = db.getAGBarcodeDetails(barcode)
+            ag_login_id = db.search_kits(details['ag_kit_id'])[0]
             site_sampled = db.human_sites
             environment_sampled = db.general_sites
-            logins = db.getAGKitsByLogin()
+            participants = db.getHumanParticipants(ag_login_id)
```
wasade: what about updating non-human barcodes?

knimin/handlers/ag_edit_barcode.py
```diff
@@ -40,7 +43,8 @@ def post(self):
                         sites_sampled=None, details=None,
                         environments_sampled=None,
                         logins=None, currentuser=self.current_user)
-        except:
+        except Exception as e:
+            print str(e)
```
wasade: Could the print be removed please?

knimin/lib/data_access.py
```diff
@@ -1588,6 +1588,26 @@ def updateAGKit(self, ag_kit_id, supplied_kit_id, kit_password,
...
+        # Get survey ID for participant if needed
+        if participant_name:
+            ag_login_id = self.search_kits(ag_kit_id)[0]
+            sql = """SELECT survey_id
+                     FROM ag_login_surveys
+                     WHERE participant_name = %s AND ag_login_id = %s"""
+            survey_id = self._con.execute_fetchone(
+                sql, [participant_name, ag_login_id])[0]
+        else:
+            survey_id = None
```
wasade: Doesn't this branch imply there is nothing to updae?

knimin/lib/data_access.py
```diff
@@ -1588,6 +1588,26 @@ def updateAGKit(self, ag_kit_id, supplied_kit_id, kit_password,
...
+            participant_name = None
+
+        # convert empty strings to None for DB consistency
+        site_sampled = site_sampled if site_sampled else None
```
wasade: these can be done with `site_sampled = site_sampled or None`

knimin/lib/tests/test_data_access.py
```diff
@@ -60,6 +60,16 @@ def test_get_unconsented(self):
         exp = ['000014660', datetime.date(2015, 3, 25), 'REMOVED']
         self.assertEqual(obs[0], exp)
 
+    def test_search_kits(self):
```
wasade: why was a test for `search_kits` added when it doesn't look like the code was modified? Glad to have more tests, but would have assumed they would be for the modified methods

knimin/lib/data_access.py
```diff
@@ -1588,6 +1588,26 @@ def updateAGKit(self, ag_kit_id, supplied_kit_id, kit_password,
...
+        if participant_name:
+            ag_login_id = self.search_kits(ag_kit_id)[0]
+            sql = """SELECT survey_id
+                     FROM ag_login_surveys
+                     WHERE participant_name = %s AND ag_login_id = %s"""
+            survey_id = self._con.execute_fetchone(
+                sql, [participant_name, ag_login_id])[0]
+        else:
+            survey_id = None
+            participant_name = None
```
wasade: On closer look, my read is that this would result in an exception in the template?

wasade:
Few comments

knimin/handlers/ag_edit_barcode.py
squirrelo: I should ad pets here too, good catch.

knimin/handlers/ag_edit_barcode.py
squirrelo: I had it in to get the error into the logs, since right now it's ignored, but I can remove it if wanted.

knimin/lib/data_access.py
squirrelo: It implies the barcode is still not attached to a survey, but site, date sampled, time sampled, refunded, or some other barcode specific piece of info was updated.

knimin/lib/data_access.py
squirrelo: oh, cool. Done

knimin/lib/tests/test_data_access.py
squirrelo: It was updated to also search over ag_kit_id column.

knimin/lib/data_access.py
squirrelo: Nope, see above for explanation.

squirrelo:
Comments addressed

squirrelo:
ping @wasade @EmbrietteH

PR #60, opened 2016-05-09
Add constants info for new sites

squirrelo:
Adds pulldown constants info for torso, left leg, and right leg.

ping @EmbrietteH @ackermag


EmbrietteH:
Looks good to me, @ackermag please double check

PR #61, opened 2016-06-01
Barcode email as unicode

squirrelo:
This makes all the email strings unicode and adds tests to ensure they are rendered correctly.


josenavas:
👍

PR #63, opened 2016-06-06
Auto set results ready

squirrelo:
This adds a button to the pulldown page that will auto-update the results ready column in the DB and send emails to users with new results. This will allow us to simply click a button to update the results ready status instead of manual editing of the DB and sending emails.


josenavas:
@squirrelo there are merge conflicts in here. Just to clarify - the button added here is just a button that when clicked checks the filesystem to detect which new barcodes have results and sends an email to the participants? i.e. there isn't anything that needs to be provided to the system - correct?

squirrelo:
Correct. It also updates the database to show results ready for the new barcodes.

josenavas:
Awesome - let me know when fixed and I will review.

squirrelo:
merge conflict fixed

knimin/lib/data_access.py
```diff
@@ -2172,6 +2176,83 @@ def get_barcode_info_by_kit_id(self, ag_kit_id):
...
+            # No new barcodes, so no emails to send
+            return
+
+        bc_sql = """UPDATE ag.ag_kit_barcodes
+                 SET date_of_last_email = '{0}'
+                 WHERE barcode IN %s""".format(datetime.now())
+        subject = "Your American/British Gut results are ready"
+        message = (
+            "Good afternoon American & British Gut participants!\n\n"
+            "We are pleased to let you know that your results are  now "
```
josenavas: Extra space between `are` and `now`

knimin/lib/data_access.py
```diff
@@ -2172,6 +2176,83 @@ def get_barcode_info_by_kit_id(self, ag_kit_id):
...
+            "available. You may access them by signing onto "
+            "microbio.me/americangut or microbio.me/britishgut. If you have "
+            "forgotten your login credentials, you may retrieve them using "
+            "the \"Forgot kit ID/password\" functions.\n\n"
+            "We thank you for being a part of the project. While we emphasize "
+            "getting results back to you, the participant, we and the broader "
+            "American/British Gutscientific collaborative network are "
```
josenavas: Add space between Gut and scientifix

knimin/lib/tests/test_mail.py
```diff
@@ -0,0 +1,32 @@
+from unittest import TestCase, main
+
+from knimin.lib.mail import send_email
+
+
+class TestEmail(TestCase):
+    def test_send_email_basic(self):
```
josenavas: Does this means that the system will send an email to `'americangut@gmail.com'` every time that the tests are executed?

josenavas:
Minor comments and a question

knimin/lib/tests/test_mail.py
squirrelo: No, it just gets the email MIME object and checks it's set up correctly. It does not actually send the email, as we can't be sure an email server is set up.

squirrelo:
Comments addressed

PR #64, opened 2016-06-07
Pulldown compliance and env pulldown

squirrelo:
This updates the pulldown code to comply with the new metadata standards. It also adds pulldown capability for environmental samples, which required updates to barcode_util so the environmental barcodes now show as good to sequence instead of a special case.

Please pay special attention to the constants set here for the pulldown. If there are more appropriate values, please advise.


josenavas:
After my last discussion with @ackermag I don't think the values specified here are correct.
@ackermag let me know if you need help going over the changes in here and we can meet up either tomorrow at ~ noon or next week.

ackermag:
I will meet with you tomorrow...

josenavas:
@ackermag and I received this - will merge once tests pass

squirrelo:
Tests are passing

PR #65, opened 2016-06-10
Add logging filepath to config

squirrelo:
This is a small PR that adds the log filepath to the config, exactly like in american_gut_web.


josenavas:
Code looks good, but the tests are failing.

squirrelo:
tests passing now

PR #68, opened 2016-06-16
Convert Crohn's disease question for pulldown

squirrelo:
This takes the answers to the new, finer-grained Crohns question and converts it to also be in line with the original rounds 1-15 question answers, so both metadata columns are in sync.

### DIFF
--- a/knimin/lib/data_access.py
+++ b/knimin/lib/data_access.py
@@ -864,6 +864,16 @@ class KniminAccess(object):
                     bc_info['ag_login_id'] + participant_name).hexdigest()
                 md[1][barcode]['PUBLIC'] = 'Yes'
 
+                # Convert finer grained IBD to coarser grained
+                ibd = md[1][barcode]['IBD_DIAGNOSIS_REFINED']
+                if ibd != 'Unspecified':
+                    if ibd in {"Ileal Crohn's Disease",
+                               "Colonic Crohn's Disease",
+                               "Ileal and Colonic Crohn's Disease"}:
+                        md[1][barcode]['IBD_DIAGNOSIS'] = "Crohn's Disease"
+                    elif ibd == 'Ulcerative colitis':
+                        md[1][barcode]['IBD_DIAGNOSIS'] = 'Ulcerative colitis'
+
                 # Add categorization columns
                 md[1][barcode]['ALCOHOL_CONSUMPTION'] = categorize_etoh(
                     md[1][barcode]['ALCOHOL_FREQUENCY'])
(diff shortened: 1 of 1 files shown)


josenavas:
@squirrelo this is not fixing the issue. We need that question and of course is not OK to change the results like this - it is important to maintain the differences.

josenavas:
Oh - NVM I completely misunderstood what was going on here

knimin/lib/data_access.py
```diff
@@ -864,6 +864,17 @@ def format_survey_data(self, md, external_surveys=None, full=False):  # noqa
...
                 md[1][barcode]['PUBLIC'] = 'Yes'
 
+                # Convert finer grained IBD to coarser grained
+                ibd = md[1][barcode].get('IBD_DIAGNOSIS_REFINED',
+                                         'Unspecified')
+                if ibd != 'Unspecified':
+                    if ibd in {"Ileal Crohn's Disease",
+                               "Colonic Crohn's Disease",
+                               "Ileal and Colonic Crohn's Disease"}:
+                        md[1][barcode]['IBD_DIAGNOSIS'] = "Crohn's disease"
```
josenavas: I understand that you want to make sure that both columns are in sync. But before changing the results shouldn't you check that the previous value is 'Unspecified'? If the participant already provided an answer to the previous question we should maintain whatever they said - correct?

squirrelo: The only way that will happen is if a person goes and edits their old survey. Since they have edited it, that means they are updating the info and therefore negating the first answer, so this is safe.

josenavas: @EmbrietteH if you agree with this, it is good to go

knimin/lib/data_access.py
```diff
@@ -864,6 +864,17 @@ def format_survey_data(self, md, external_surveys=None, full=False):  # noqa
...
+                    elif ibd == 'Ulcerative colitis':
+                        md[1][barcode]['IBD_DIAGNOSIS'] = 'Ulcerative colitis'
+
```
EmbrietteH: How is the answer of "microcolitis" treated? Here it looks like it's just not grouped into either Crohn's or UC.

squirrelo: It is fits neither category of the first question, so it is undefined.

EmbrietteH: Have @JWDebelius and @ackermag approved this non-categorization? Just want to make sure it is physiologically correct...

ackermag: Why have it as a category if it is undefined?  How will it be dented in pull-down?

ackermag: sorry denoted.

ackermag: It is a histologic diagnosis and pretty distinct.

squirrelo: This is converting the new more fine grained question to also fill the old "UC or Crohns" question. Therefore it  will be undefined in the old question since it's not actually part of that old question.

PR #69, opened 2016-06-20
Barcode util failure reason

squirrelo:
This shows failure reasons on the barcodes scan page when the barcode is not registered.


josenavas:
Code looks good - but tests are failing.

squirrelo:
test updated for the new text.

PR #70, opened 2016-06-27
Fix for removal of participant_name from ag_kit_barcodes

squirrelo:
This updates labadmin to comply with the database change in american gut centralizing the participant_name column.

### DIFF
--- a/knimin/lib/tests/test_data_access.py
+++ b/knimin/lib/tests/test_data_access.py
@@ -121,6 +121,99 @@ class TestDataAccess(TestCase):
                [6, 'Search'], [7, 'Admin']]
         self.assertEqual(obs, exp)
 
+    def test_participant_names(self):
+        obs = db.participant_names()
+        self.assertEqual(len(obs), 8237)
+        self.assertEqual(obs[0], ['000027561', 'REMOVED-0'])
+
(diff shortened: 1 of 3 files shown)


knimin/lib/tests/test_data_access.py
```diff
@@ -121,6 +121,99 @@ def test_get_access_levels(self):
...
+        obs = db.participant_names()
+        self.assertEqual(len(obs), 8237)
+        self.assertEqual(obs[0], ['000027561', 'REMOVED-0'])
+
+    def test_search_barcodes(self):
+        obs = db.search_barcodes('000001124')
+        self.assertEqual(obs, ['d8592c74-7c27-2135-e040-8a80115d6401'])
+
+        obs = db.search_barcodes('REMOVED-8')
+        exp = ['d8592c74-9491-2135-e040-8a80115d6401',
```
josenavas: There are a few values repeated in here. Does that mean that there is a distinct missing in a query?

squirrelo: yup, distinct added.

squirrelo:
Comment addressed

josenavas:
👍

PR #73, opened 2016-06-30
Redirect if admin removes own admin privileges

squirrelo:


josenavas:
I'm guessing there are no tests for this?

PR #78, opened 2016-07-09
Added installation instructions to README

tanaes:
self explanatory 🍆


mortonjt:
👍 I think this looks good.

PR #81, opened 2016-07-28
Fixing pulldown

josenavas:
While testing on the test environment we realized that the columns not allowed by EBI where added to the pulldown. Although those columns were always containing only the value "Unspecified" (i.e. they did not contain real data), those columns should not be there.

@wasade, are you able to take a quick review?


EmbrietteH:
👍 ping @wasade

PR #84, opened 2016-08-17
TST: Add tests for squash barcodes

ElDeveloper:
These tests are minimal, and I think it's as far as we can go, but I'm open to other suggestions.


josenavas:
👍

mortonjt:
Looks like there is 1 failing test and some flake8 errors

ElDeveloper:
Thanks for the reviews, just fixed the flake8 errors.

mortonjt:
👍

PR #93, opened 2016-09-13
Scan issues

wasade:
@EmbrietteH, these should resolve the current issues we're seeing


wasade:
@EmbrietteH, this should cover the email_type issue. Working on generating the necessary state for an alternative survey to trigger the other one

EmbrietteH:
👍

PR #98, opened 2016-09-16
Plate mapper kiwi checkpoint

qiyunzhu:


qiyunzhu:
Dear Jose and Daniel,

Thank you for your patience. My first pull request of the plate mapper from the labadmin end is online for your review.

Based on the previous discussion between Jose and me, I only kept a small amount of edits with highest priority, while removed all other functions such as GUI and data migration. Therefore only two files were modified: data_access.py and test_data_access.py.

The major functionality I added are the database read/write operations of plate attributes and plate-to-sample map. They were achieved in four functions: get_plate_info, set_plate_info, get_plate_map, and set_plate_map.

In addition there is a fifth function: get_plate_type. In my previous email I explained an idea of uniformizing field option retrieval operations. In this version I still kept the conventional way of getting plate types. But we can think of using the new idea later.

I wrote unit tests for all those five functions. It wasn't easy, as it involves database read/write operations. I had to combine the tests for the four plate functions into one single test function, since a plate has to be created prior to the creation of the plate map. Meanwhile, I also added three additional functions: _add_new_samples, _delete_new_samples, and _delete_new_plate, to serve as setup and teardown functionalities for the test of the four functions.

Thank you for reviewing my code! Looking forward to hearing your valuable feedbacks!

Best,
Qiyun

qiyunzhu:
Please note that, despite that I made efforts to restore the database to its original status after unit tests, it won't be perfectly identical. The plate ID as a sequence will irreversibly ++ after removing the created plate.

PR #99, opened 2016-09-16
Plate mapper kiwi checkpoint

qiyunzhu:
Dear Jose and Daniel,

Thank you for your patience. My first pull request of the plate mapper from the labadmin end is online for your review.

Based on the previous discussion between Jose and me, I only kept a small amount of edits with highest priority, while removed all other functions such as GUI and data migration. Therefore only two files were modified: data_access.py and test_data_access.py.

The major functionality I added are the database read/write operations of plate attributes and plate-to-sample map. They were achieved in four functions: get_plate_info, set_plate_info, get_plate_map, and set_plate_map.

In addition there is a fifth function: get_plate_type. In my previous email I explained an idea of uniformizing field option retrieval operations. In this version I still kept the conventional way of getting plate types. But we can think of using the new idea later.

I wrote unit tests for all those five functions. It wasn't easy, as it involves database read/write operations. I had to combine the tests for the four plate functions into one single test function, since a plate has to be created prior to the creation of the plate map. Meanwhile, I also added three additional functions: _add_new_samples, _delete_new_samples, and _delete_new_plate, to serve as setup and teardown functionalities for the test of the four functions.

Thank you for reviewing my code! Looking forward to hearing your valuable feedbacks!

Best,
Qiyun


qiyunzhu:
Please note that, despite that I made efforts to restore the database to its original status after unit tests, it won't be perfectly identical. The plate ID as a sequence will irreversibly ++ after removing the created plate.

knimin/lib/data_access.py
```diff
@@ -2399,6 +2399,134 @@ def getProjectNames(self):
...
+
+        Returns
+        -------
+        dict
+            Plate attributes
+        """
+        sql = """SELECT * FROM pm.plate WHERE plate_id = %s"""
+        return self._con.execute_fetchdict(sql, [plate_id])[0]
+
+    def set_plate_info(self, plate_id, plate_info):
```
josenavas: Split this function in 2: "create_plate" and "set_plate_info", it will make the code more readable. The "set_plate_info" will not need to return the plate id.

knimin/lib/data_access.py
```diff
@@ -2399,6 +2399,134 @@ def getProjectNames(self):
...
+        Returns
+        -------
+        int
+            Plate id of the created / modified plate
+        """
+        keys = ', '.join([str(x) for x in plate_info.keys()])
+        values = ', '.join(['\'' + str(x) + '\'' for x in plate_info.values()])
+        sql = ''
+        if plate_id == 0:  # Create new plate
+            sql = 'INSERT INTO pm.plate (' + keys + ') VALUES (' + values + ')'
```
josenavas: This is not SQL safe - this can generate really confusing errors. When we are working with SQL we really want to control what things that we are inserting. Also, this way of passing the values to the SQL is really dangerous and highly discouraged. Review the [SQL coding guidelines in QIITA](https://github.com/biocore/qiita/blob/master/CONTRIBUTING.md#sql-coding-guidelines) and specially point 4 for the argument of why there are problems of doing this in this way.

knimin/lib/data_access.py
```diff
@@ -2399,6 +2399,134 @@ def getProjectNames(self):
...
+        Returns
+        -------
+        list of tuple of (int, int, str)
+            Column id, row id, sample id
+        """
+        sql = """SELECT col, row, sample_id FROM pm.plate_sample
+                 WHERE plate_id = %s"""
+        return self._con.execute_fetchall(sql, [plate_id])
+
+    def set_plate_map(self, plate_id, plate_map):
```
josenavas: This functions is doing multiple different operations, that is confusing, and given the amount of SQL queries that can happen (96 well plate * 2 queries per well = 192 SQL queries) can make this function really slow.

If the idea of set_plate_map is to change all the contents of the plate, I think it is easier to remove the old data and insert the new data. This will be only 2 queries. However, in order to do that, we need transactions, that are currently not supported by the SQL connection handler in this repo.

@wasade any thoughts?

knimin/lib/data_access.py
```diff
@@ -2399,6 +2399,134 @@ def getProjectNames(self):
...
+                sql_args = [plate_id, col, row]
+                sql = """DELETE FROM pm.plate_sample WHERE plate_id = %s
+                         AND col = %s AND row = %s RETURNING sample_id"""
+                self._con.execute(sql, sql_args)
+                ndel += 1
+        return (nadd, nmod, ndel)
+
+    def get_plate_type(self, plate_id):
+        """Gets plate type attributes
+        Parameters
```
josenavas: Add a blank line

knimin/lib/data_access.py
```diff
@@ -2399,6 +2399,134 @@ def getProjectNames(self):
...
+            >0: plate type of an exisiting plate by id
+
+        Returns
+        -------
+        dict
+            Attributes of a plate type, currently including: plate_type_id,
+            name, cols, rows, notes
+        """
+        plate_type_id = 1
+        if plate_id > 0:
```
josenavas: I don't understand why this if is needed? Can you elaborate please? I have the feeling that this can be reduced to a single SQL query:

```python
sql = """SELECT pt.* FROM pm.plate_type pt JOIN pm.plate USING (plate_type_id) WHERE plate_id = %s"""
return self._con..execute_fetchdict(sql, [plate_id])[0]
```

knimin/lib/data_access.py
```diff
@@ -2425,3 +2553,38 @@ def _revert_ready(self, barcodes):
                  SET results_ready = NULL
                  WHERE barcode IN %s"""
         self._con.execute(sql, [tuple(barcodes)])
+
+    def _add_new_samples(self, sample_ids):
```
josenavas: Call this "create_sample" and add tests for it, it will be a function that we will be using at some point (not only for tests) so since you've already written it we can make it available

knimin/lib/data_access.py
```diff
@@ -2425,3 +2553,38 @@ def _revert_ready(self, barcodes):
...
+
+        Parameters
+        ----------
+        sample_ids : list of str
+            Sample ids to create
+        """
+        sql = """INSERT INTO pm.sample (sample_id) VALUES (%s)"""
+        self._con.executemany(sql, [[x] for x in sample_ids])
+
+    def _delete_new_samples(self, sample_ids):
```
josenavas: "delete_samples" and same comment as above

knimin/lib/data_access.py
```diff
@@ -2425,3 +2553,38 @@ def _revert_ready(self, barcodes):
...
+
+        Parameters
+        ----------
+        sample_ids : list of str
+            Sample ids to delete
+        """
+        sql = """DELETE FROM pm.sample WHERE sample_id = %s"""
+        self._con.executemany(sql, [[x] for x in sample_ids])
+
+    def _delete_new_plate(self, plate_id):
```
josenavas: "delete_plate" and same comment as above

knimin/lib/tests/test_data_access.py
```diff
@@ -297,6 +297,99 @@ def test_get_ag_barcode_details(self):
             del(obs[key]['registered_on'])
         self.assertEqual(obs, exp)
 
+    def test_plate_operations(self):
```
josenavas: Understand your comment about being "dependent", but usually unit test test small functionality. I think you should split up the tests for each of them, and if you need to create another plate your one of your other tests, you can create it. It may seem some code duplication, but having unit tests that actually test units are really helpful to debug later.

knimin/lib/tests/test_data_access.py
```diff
@@ -297,6 +297,99 @@ def test_get_ag_barcode_details(self):
...
+        plate_info = {
+            'processing_robot_id': 2,  # RIKE
+            'tm300_8_tool_id': 3,  # 109375A
+        }
+        obs = db.set_plate_info(plate_id, plate_info)
+        exp = plate_id
+        self.assertEqual(obs, exp)
+
+        # Test get_plate_info:
+        exp = {
```
josenavas: You probably don't want to return IDs, but rather change them internally in the function, or return it as a tuple with (id, name). Take a look at the JOIN that I used above. You'll need a lot of joins when working with SQL, so it is good if you get familiarized with them.

qiyunzhu:
@josenavas Thank you for your very in-time code review! I will read and think about your comments.

knimin/lib/data_access.py
qiyunzhu: I see. Thanks for pointing me to the documentation! I originally used sql_args, but found that it didn't work in this case. What I tried is like:
```
sql = """INSERT INTO pm.plate (%s) VALUES (%s)"""
self._con.execute_fetchone(sql, [keys, values])
```
But the string was not formatted correctly. I don't know why. Is there an easy way to do this?

knimin/lib/data_access.py
qiyunzhu: There was a reason in it: that some plate table columns are not null (e.g., "email"). It is not possible to create a plate without setting these fields.

knimin/lib/data_access.py
qiyunzhu: You are right! I didn't know this trick before you inspired me.

knimin/lib/data_access.py
qiyunzhu: Sure!

knimin/lib/data_access.py
qiyunzhu: same as above.

knimin/lib/data_access.py
qiyunzhu: same as above.

knimin/lib/tests/test_data_access.py
qiyunzhu: I see. If I put them independently, these functions will call each other in order to complete a test (e.g., create_plate -> set_plate_info -> get_plate_info -> delete_plate). Will this be a problem?

knimin/lib/tests/test_data_access.py
qiyunzhu: I am not fully clear with this comment. I guess that you mean that the get_plate_info function should return everything else but plate_id? This sounds hard to do. To return everything we just need SELECT * FROM, but to return everything but one thing, what should I do?

knimin/lib/data_access.py
qiyunzhu: You comment makes sense to me! Is there an easy single SQL command to insert multiple rows all at once Also waiting for @wasade 's comment.

wasade: I agree that it would be good to perform these operations in a single query or transaction. I'm less concerned about performance given the scope of this repo, and more concerned about atomicity. 

Can this be expressed entirely as a single insert? I guess I'm not clear as to why an update or delete would be performed. It also doesn't seem like the result of `RETURNING` is used?

@josenavas, I don't know if TRN is or is not in the repo. That is of course one way to deal with transactions.

knimin/lib/data_access.py
```diff
@@ -2399,6 +2399,134 @@ def getProjectNames(self):
...
+        plate_info : dict
+            Plate attributes
+
+        Returns
+        -------
+        int
+            Plate id of the created / modified plate
+        """
+        keys = ', '.join([str(x) for x in plate_info.keys()])
+        values = ', '.join(['\'' + str(x) + '\'' for x in plate_info.values()])
```
wasade: @josenavas, psycopg2 will escape things if asked, right? (blanking on the means to do this right now)

knimin/lib/data_access.py
```diff
@@ -2425,3 +2553,38 @@ def _revert_ready(self, barcodes):
...
+
+    def _delete_new_samples(self, sample_ids):
+        """Test helper to delete the samples that were just created
+
+        Parameters
+        ----------
+        sample_ids : list of str
+            Sample ids to delete
+        """
+        sql = """DELETE FROM pm.sample WHERE sample_id = %s"""
```
wasade: should this `DELETE` be restricted to a particular plate? For instance, what about the scenario where the same sample is run multiple times? One risk is that, right now this method is used as a test helper, but it may get reused inadvertently in the future.

knimin/lib/data_access.py
qiyunzhu: @wasade You are right that the RETURNING is not used. It is there for historical reason. I will remove them.

knimin/lib/data_access.py
qiyunzhu: @wasade The scenario you proposed is possible and has been taken consideration of. Therefore when the end user deletes a plate, the samples will still be in the database. The current function is dedicated for deleting samples though.

knimin/lib/data_access.py
qiyunzhu: Never mind. I think I figured out a solution, according to this [post](http://stackoverflow.com/questions/28916892/passing-a-list-of-column-names-and-values-to-psycopg).

knimin/lib/data_access.py
josenavas: I think that what you're saying actually reinforces my point. The "create_plate" and "set_plate_info" are two different operations that need to execute different checks.

I know the database is doing those checks for you, but debugging with errors from SQL has been hard in the past.

This actually exposed the need for a better API. Instead of accepting a dictionary with all possible data, change the API to be verbose, and specifically ask for the parameters that are needed, while letting the other one nullable. An example of this idea is in the artifact creation in Qiita: https://github.com/biocore/qiita/blob/master/qiita_db/artifact.py#L159.

Then you can create the "set" function to change the parameters that make sense to change, not all of them are going to be 'setable'. For example, the user that creates the plate.

You can read that link that I sent above regarding the artifact object in Qiita. It will help you to understand how we use python to limit what can be done in the DB and provide useful errors.

knimin/lib/data_access.py
josenavas: The column names are not supported to be passed through parameters, only the values. That is intended by the psycopg2 developers. The way of doing this specific call will be something along the lines of:
```python
sql = "INSERT INTO pm.plate ({}) VALUES ({})".format(keys, ','.join(['%s'] * len(plate_info)))
self._con.execute_fetchone(sql, values)
```

This way psycopg2 will do the conversion for you - and will be safer.

knimin/lib/data_access.py
josenavas: The TRN object is not in use in the project, this is using the old SQLConnectionHandler... ugh. Changing to the new TRN object would be nice, probably adding it as a goal for the code-sprint will be good?

knimin/lib/tests/test_data_access.py
josenavas: Nope, that is ok. If there is a bug in any of them, the specific test will fail, so we can easily identify which one is the source of the problem.

knimin/lib/data_access.py
josenavas: Correct - changing it as I suggested above/below (new github reviews - not sure where this will end  up) will fix it.

knimin/lib/data_access.py
josenavas: @qiyunzhu Note that this function will fail if the sample is attached to a plate - thus we need to perform some validation before "blindly" remove the sample. We probably don't want to let the user remove the sample if it is attached to a plate.

josenavas:
It's me, or the new Github reviews are kinda annoying? lol

knimin/lib/tests/test_data_access.py
qiyunzhu: Got it!

knimin/lib/data_access.py
qiyunzhu: Thanks. Let me read it. Doesn't sound easy to me at this moment.

knimin/lib/data_access.py
qiyunzhu: How about this (following my last comment):
```
from psycopg2.extensions import AsIs
sql_args = [AsIs(', '.join(plate_info.keys())), tuple(plate_info.values())]
sql = """INSERT INTO pm.plate (%s) VALUES %s"""
return self._con.execute_fetchone(sql, sql_args)[0]
```
This avoids the need of Python string formatting operations.

knimin/lib/tests/test_data_access.py
qiyunzhu: I understand your rationale. However, I am thinking that the front end will have to do additional calls to retrieve actual values anyway, because the drop-down menus needs items. Therefore returning ids and letting the front end find and display actual values seems to be a better solution. Maybe there are other ways to do this?

knimin/lib/data_access.py
```diff
@@ -2399,64 +2400,205 @@ def getProjectNames(self):
         sql = """SELECT project FROM project"""
         return [x[0] for x in self._con.execute_fetchall(sql)]
 
+<<<<<<< HEAD
```
wasade: it looks like there were conflicts which got committed?

knimin/lib/data_access.py
```diff
@@ -2399,64 +2400,205 @@ def getProjectNames(self):
...
+        Returns
+        -------
+        bool
+            True if successful
+        """
+        sql = """INSERT INTO pm.sample (sample_id) VALUES (%s)"""
+        self._con.executemany(sql, [[x] for x in sample_ids])
+        return True
+
+    def delete_sample(self, sample_ids):
```
wasade: isn't there a method `deleteSample`? it may be good to make it clear on the method names that these methods are associated with the plate mepper

knimin/lib/data_access.py
```diff
@@ -2399,64 +2400,205 @@ def getProjectNames(self):
...
-            Plate id
+        sample_ids : list of str
+            Sample ids to delete
 
         Returns
         -------
-        dict
-            Plate attributes
+        bool
+            True if successful
```
wasade: will this ever return false?

knimin/lib/data_access.py
```diff
@@ -2399,64 +2400,205 @@ def getProjectNames(self):
...
+        sql = """DELETE FROM pm.sample WHERE sample_id = %s"""
+        self._con.executemany(sql, [[x] for x in sample_ids])
+        return True
+
+    def create_plate(self, name, email, plate_type):
+        """Creates a new plate and sets values for mandatory fields
+
+        Parameters
+        ----------
+        name : str
```
wasade: would it be possible to add in descriptions of the parameters?

knimin/lib/data_access.py
```diff
@@ -2399,64 +2400,205 @@ def getProjectNames(self):
...
+        ----------
+        plate_ids : list of int
+            Plate ids to delete
+
+        Returns
+        -------
+        int
+            Number of plates deleted
+        """
+        sql = """DELETE FROM pm.plate_sample WHERE plate_id = %s"""
```
wasade: can these statements be performed within a transaction to ensure atomicity?

knimin/lib/data_access.py
```diff
@@ -2399,64 +2400,205 @@ def getProjectNames(self):
...
-            sql = 'UPDATE pm.plate SET (' + keys + ') = (' + values + ')'
-            sql += ' WHERE plate_id = ' + str(plate_id)
-        sql += ' RETURNING plate_id'
-        return self._con.execute_fetchone(sql)[0]
+        sql = """UPDATE pm.plate
+                 SET name = %s,
+                     template_id = (SELECT template_id
+                        FROM pm.template WHERE name = %s),
+                     linker_seq = %s,
+                     extraction_kit_lot_id = (SELECT extraction_kit_lot_id
```
wasade: shouldn't these be joins instead of nested selects?

knimin/lib/data_access.py
```diff
@@ -2520,13 +2678,65 @@ def get_plate_type(self, plate_id):
...
-        return self._con.execute_fetchdict(sql, [plate_type_id])[0]
+            sql = """SELECT pm.plate_type.* FROM pm.plate_type JOIN pm.plate
+                     USING (plate_type_id) WHERE plate_id = %s"""
+            sql_args = [plate_id]
+        else:
+            sql = """SELECT * FROM pm.plate_type WHERE plate_type_id = %s"""
+            sql_args = [1]
+        return self._con.execute_fetchdict(sql, sql_args)[0]
+
+<<<<<<< HEAD
```
wasade: more conflict bits

knimin/lib/data_access.py
qiyunzhu: I found it too and I am fixing it.

knimin/lib/data_access.py
qiyunzhu: No, it will only raise error if the SQL command doesn't run thru.

qiyunzhu:
Dear all,

Here comes my new trial. Now in data_access.py I have 12 new functions. They are all unit-tested. They are:

**get_id_by_name** - inspired Qiita's [convert_to_id](https://github.com/biocore/qiita/blob/master/qiita_db/util.py#L894) as kindly pointed to by @josenavas . I actually didn't use this function in the current code. Maybe useful later.

**create_sample**, **delete_sample** - As the names suggest. Currently they only create sample_id's. Later we may add columns to the sample table and enrich these functions.

**create_plate**, **delete_plate** - Based on @josenavas 's suggestion, create_plate is separated from set_plate_info. Now create_plate only sets values to mandatory columns (name, email, plate type).

**set_plate_info**, **get_plate_info** - Based on @josenavas 's suggestion, the real IDs of options now will not be sent to the interface. The users only see the names. I still have plate_id sent to the user, because I think plate_id makes much sense to the user as an identifier. In contrast, plate names (which may read "Kiwi's favorite plate #1") are harder to track by the user.

**set_plate_map**, **get_plate_map** - I didn't change the function for the time. @josenavas suggested that 192 SQL queries may impair performance. Is there a good solution to reduce SQL queries?

**get_plate_type** - Modified based on comments.

**get_plate_count** - Counts the total number of plates (to be displayed on the page). This function seems to overlap the existing function db.getAGStats somehow.

**get_plate_list** - This is a new function. I use it to return a list of plates with their most basic information. This will then be rendered on the screen is a "plate list" page.

knimin/lib/data_access.py
qiyunzhu: Hi @wasade , I am impressed that you reviewed the code so timely! I was still checking the code and writing comments. Sorry for the conflict. Now it is done and re-PR'ed. I am not very familiar with the GitHub review system, so I am not sure if you can see this comment. Can you please reply to confirm if you see it?

knimin/lib/data_access.py
qiyunzhu: I didn't see `deleteSample`. But I think it makes sense if we make the naming of the functions more unique to plate mapper (e.g., add prefix `pm_`?). @josenavas what do you think?

knimin/lib/data_access.py
qiyunzhu: They won't return False. They will only raise error if the SQL command doesn't run thru.

knimin/lib/data_access.py
qiyunzhu: These are identical to database column names. Do we still write descriptions?

knimin/lib/data_access.py
qiyunzhu: I don't know how JOIN works in an UPDATE command. @wasade can you give me an example?

knimin/lib/data_access.py
qiyunzhu: I wrote some comments in the master comment in front of the entire PR.

knimin/lib/data_access.py
qiyunzhu: Resolved.

knimin/lib/data_access.py
wasade: oh, i see, was looking at this wrong. i think you can join all these tables below by `name` and take a projection of the various IDs to pull out? approximately like this http://stackoverflow.com/a/6258586

knimin/lib/data_access.py
wasade: Not sure I understand. What about just issuing begin/commit, and just issue the statements direct against the connection object since TRN isn't in this repo? cc @josenavas

mortonjt:
Looks like you have a flake8 error in last travis build.  

You can probably just delete knimin/lib/data_access.py:18:1

qiyunzhu:
Dear all, I submitted a new pull request (#99).

I tried to make the code simple as possible for review. There are only 8 new functions:
create_sample, delete_sample,
create_sample_plate, delete_sample_plate,
set_sample_plate_info, get_sample_plate_info,
set_sample_plate_layout, get_sample_plate_layout,
plus _execute_sql which is used to pre-populate the tables in setup stage

The codes are basically some re-writings from my last PR (also #99). Some extra functions were intensionally removed from the current version.

Please not that this PR is compatible with the merged amgut PR [#620](https://github.com/biocore/american-gut-web/pull/620), which still has a bug (see my comment). Therefore PLEASE DO NOT MERGE THE CURRENT PR.

knimin/lib/data_access.py
```diff
@@ -2425,3 +2597,7 @@ def _revert_ready(self, barcodes):
                  SET results_ready = NULL
                  WHERE barcode IN %s"""
         self._con.execute(sql, [tuple(barcodes)])
+
+    def _execute_sql(self, sql, sql_args=None):
```
josenavas: I don't think this function is needed. Note that you can change all the instances of this function by:
```python
db._con.execute(sql, sql_args)
```

knimin/lib/tests/test_data_access.py
```diff
@@ -19,10 +19,64 @@ def setUp(self):
             db.add_external_survey('Vioscreen', 'FFQ', 'http://vioscreen.com')
         except ValueError:
             pass
+        # Populate some field options
+        sql = """INSERT INTO pm.plate_type (name, cols, rows, notes)
+                 VALUES ('96-well', 12, 8, 'Standard 96-well plate')"""
+        db._execute_sql(sql)
+        sql = """INSERT INTO pm.extraction_robot (name) VALUES ('HOWE_KF1')"""
```
josenavas: You can insert multiple values in the same table by using a single INSERT command:
```sql
INSERT INTO pm.extraction_robot (name)
VALUES ('HOWE_KF1'), ('HOWE_KF2'), ('HOWE_KF3'), ('HOWE_KF4')
```

knimin/lib/tests/test_data_access.py
```diff
@@ -243,60 +297,138 @@ def test_getAnimalParticipantsNotPresent(self):
...
+               'survey_id': '084532330aca5885',
+               'other': 'N',
+               'sample_barcode_file_md5': None
+               }}
         for key in obs:
             del(obs[key]['registered_on'])
         self.assertEqual(obs, exp)
 
+    def test_create_sample(self):
+        """"""
```
josenavas: If you are not adding any comment, just remove this line (in all new tests).

knimin/lib/data_access.py
```diff
@@ -2399,6 +2399,178 @@ def getProjectNames(self):
...
+            Sample IDs to delete
+
+        Returns
+        -------
+        bool
+            True if successful
+        """
+        sql = """DELETE FROM pm.sample WHERE sample_id = %s"""
+        self._con.executemany(sql, [[x] for x in sample_ids])
+        return True
```
josenavas: Remove the return `True` from all these functions, it is not providing any useful information. If the function doesn't run correctly it will raise an error that it will be handled somewhere else.

knimin/lib/data_access.py
```diff
@@ -2399,6 +2399,178 @@ def getProjectNames(self):
         sql = """SELECT project FROM project"""
         return [x[0] for x in self._con.execute_fetchall(sql)]
 
+    def create_sample(self, sample_ids):
```
josenavas: What if the sample already exists? The DB will raise an error, but it may be useful to define exactly which is the behavior that we want to happen.

Also, the samples will always belong to a project. How do you associate them to a project?

knimin/lib/data_access.py
```diff
@@ -2399,6 +2399,178 @@ def getProjectNames(self):
...
+        Returns
+        -------
+        bool
+            True if successful
+        """
+        sql = """INSERT INTO pm.sample (sample_id) VALUES (%s)"""
+        self._con.executemany(sql, [[x] for x in sample_ids])
+        return True
+
+    def delete_sample(self, sample_ids):
```
josenavas: If the sample is associated with a project and/or with a sample plate, the DB will raise an error. We should probably put some checks in place before blindly removing the sample.

Also, it looks like this function is used to delete multiple samples, so it is better to update the name accordingly: `delete_samples`

knimin/lib/data_access.py
```diff
@@ -2399,6 +2399,178 @@ def getProjectNames(self):
...
+        Returns
+        -------
+        bool
+            True if successful
+        """
+        sql = """DELETE FROM pm.sample WHERE sample_id = %s"""
+        self._con.executemany(sql, [[x] for x in sample_ids])
+        return True
+
+    def create_sample_plate(self, sample_plate_info):
```
josenavas: instead of asking for a tuple with a specific ordering, ask for the different parameters:
```python
def create_sample_plate(self, name, email, plate_type, creation_timestamp=None, notes=None):
```

knimin/lib/data_access.py
```diff
@@ -2399,6 +2399,178 @@ def getProjectNames(self):
...
+        sql = """INSERT INTO pm.sample_plate (name, email, creation_timestamp,
+                 notes, plate_type_id) VALUES (%s, %s, %s, %s,
+                 (SELECT plate_type_id FROM pm.plate_type WHERE name = %s))
+                 RETURNING sample_plate_id"""
+        sql_args = list(sample_plate_info)
+        for i in range(1, 4):
+            sql_args[i] = sql_args[i] or None
+        return self._con.execute_fetchone(sql, sql_args)[0]
+
+    def set_sample_plate_info(self, sample_plate_id, sample_plate_info):
```
josenavas: Same as above regarding parameters

knimin/lib/data_access.py
```diff
@@ -2399,6 +2399,178 @@ def getProjectNames(self):
...
+        ----------
+        sample_plate_info: tuple of (str, str, datetime, str, str)
+            name, email, creation_timestamp, notes, plate_type
+
+        Returns
+        -------
+        int
+            ID of the created sample plate
+        """
+        sql = """INSERT INTO pm.sample_plate (name, email, creation_timestamp,
```
josenavas: Can you update the SQL queries to follow the SQL coding guidelines exposed in [Qiita's contributing file](https://github.com/biocore/qiita/blob/master/CONTRIBUTING.md#sql-coding-guidelines)? Points 2 and 3 are the ones that apply here. As an example, this should read something like:
```python
sql = """INSERT INTO pm.sample_plate (name, email, creation_timestamp,
                                      notes, plate_type_id)
         VALUES (%s, %s, %s, %s, (SELECT plate_type_id
                                  FROM pm.plate_type
                                  WHERE name = %s))
         RETURNING sample_plate_id"""
```

knimin/lib/data_access.py
```diff
@@ -2399,6 +2399,178 @@ def getProjectNames(self):
...
+            ID of the sample plate to set attributes to
+        sample_plate_info: tuple of (str x2, datetime, str x2)
+            name, email, creation_timestamp, notes, plate_type
+
+        Returns
+        -------
+        bool
+            True if successful
+        """
+        sql = """UPDATE pm.sample_plate
```
josenavas: Same as above regarding SQL coding guidelines

knimin/lib/data_access.py
```diff
@@ -2399,6 +2399,178 @@ def getProjectNames(self):
...
+            name, email, creation_timestamp, notes, plate_type
+        """
+        sql = """SELECT p.name, p.email, p.creation_timestamp, p.notes,
+                    pm.plate_type.name
+                 FROM pm.sample_plate p
+                 JOIN pm.plate_type USING (plate_type_id)
+                 WHERE sample_plate_id = %s"""
+        return tuple(self._con.execute_fetchone(sql, [sample_plate_id]))
+
+    def delete_sample_plate(self, sample_plate_ids):
```
josenavas: Change to `delete_sample_plates`.

In which case are we going to delete multiple sample plates at once? I think it is safer if we only allow to remove 1 sample plate at a time (note that once it is removed is gone forever, so we want to protect a bit better for accidental deletions).

Also, this will fail in case that the SamplePlate is associated with a DNA plate.

knimin/lib/data_access.py
```diff
@@ -2399,6 +2399,178 @@ def getProjectNames(self):
...
+        """
+        sql = """DELETE FROM pm.sample_plate_layout
+                 WHERE sample_plate_id = %s"""
+        self._con.executemany(sql, [[x] for x in sample_plate_ids])
+        sql = """DELETE FROM pm.sample_plate
+                 WHERE sample_plate_id = %s"""
+        self._con.executemany(sql, [[x] for x in sample_plate_ids])
+        return True
+
+    def set_sample_plate_layout(self, sample_plate_id, sample_plate_layout):
```
josenavas: There is no checks in place that make sure that `sample_plate_layout` is actually making a valid plate layout.

knimin/lib/tests/test_data_access.py
```diff
@@ -243,60 +297,138 @@ def test_getAnimalParticipantsNotPresent(self):
...
+        sample_ids = ['test_sample_1']
+        obs = db.create_sample(sample_ids)
+        self.assertTrue(obs)
```
josenavas: This tests are not checking that the functionality is actually working. For example, this functions will make this tests to pass:
```python
def create_sample(sample_ids):
    return True

def delete_sample(sample_ids):
    return True
```

That's why the return True that you have in the functions are not meaningful. The correct tests need to check the contents of the DB.

knimin/lib/data_access.py
```diff
@@ -2399,6 +2399,178 @@ def getProjectNames(self):
...
+        ----------
+        sample_plate_ids : list of int
+            Sample plate IDs to delete
+
+        Returns
+        -------
+        bool
+            True if successful
+        """
+        sql = """DELETE FROM pm.sample_plate_layout
```
wasade: I do think we want these statements combined into a transaction. @josenavas, it's been a bit since I've looked at this with `psycopg2`, but in the absence of the `TRN` object, is the following sufficient?

```python
with self._con.cursor() as cur:
    cur.execute('BEGIN')
    # ... do stuff ...
    cur.execute('COMMIT')
```

knimin/lib/data_access.py
```diff
@@ -2399,6 +2399,178 @@ def getProjectNames(self):
...
+            Sample plate ID
+        sample_plate_layout : list of tuple of (str, int, int, str, str)
+            Sample ID, column ID, row ID, name, notes
+
+        Returns
+        -------
+        bool
+            True if successful
+        """
+        sql = """DELETE FROM pm.sample_plate_layout
```
wasade: same here w.r.t. transaction. basically, i worry about worst case scenario, either a disrupted connection, power outage, or exception between queries which could lead to state inconsistencies in the database through a partial execution of one of these methods.

knimin/lib/tests/test_data_access.py
```diff
@@ -243,60 +297,138 @@ def test_getAnimalParticipantsNotPresent(self):
...
+            ('test_sample_2', 1, 2, 'B', 'duplicate'),
+            ('test_sample_3', 1, 3, '', 'low'),
+        ]
+        obs = db.set_sample_plate_layout(spid, splayout)
+        self.assertTrue(obs)
+        db.delete_sample_plate([spid])
+        db.delete_sample(sample_ids)
+
+    def test_get_sample_plate_layout(self):
+        """"""
```
wasade: it's not necessary to have the empty docstrings

qiyunzhu:
Hi @josenavas and @wasade I am in Irvine now. Do you now have a common point of whether using transaction or other sorts of verification?

knimin/lib/tests/test_data_access.py
qiyunzhu: Thanks! Edited.

knimin/lib/tests/test_data_access.py
qiyunzhu: Got it. Edited.

knimin/lib/data_access.py
qiyunzhu: Edited.

knimin/lib/data_access.py
qiyunzhu: Got it. Will edit.

knimin/lib/data_access.py
qiyunzhu: Got it!

knimin/lib/tests/test_data_access.py
qiyunzhu: Edited.

knimin/lib/data_access.py
qiyunzhu: You are right! Since the DNA plate functionalities have not been implemented, I left this function free of DNA plate check.

I anticipate that the user can select multiple plates in the "plate list" page and delete them all at once. There will be warnings of course. Maybe let's leave the Python function multiple and later think whether the interface allows multiple or not?

knimin/lib/data_access.py
qiyunzhu: Thanks!

qiyunzhu:
Dear all,

Thanks for your valuable comments on my last attempt. I have worked out a new version. This is a very minimal increment, including only one simplest functional module: **create/edit/read/delete study**. Yet, you may find that the code is _much more complicated_ than my previous version, as I have tried my best to implement data verification @josenavas , error handling @josenavas and a very simple transaction @wasade . I hope that we can review and improve this simplest module to an acceptable status. Then I will code other modules accordingly.

Some notes: 1) even being complicated already, the code does not look for samples assigned to the relevant studies. This function will be added later once the sample module is implemented.

2) The functions still return `True` is successful. But data validation steps were added so that it raises errors if not successful. All error types were set to be `ValueError`, with meaningful error messages. They are all are tested. I anticipate that the front end will be able to read the error messages and get back to the user.

2.5) @wasade Data validation steps has `RETURNING` clauses. They do not actually use the returned variable, but merely test if it's None. `RETURNING` is to make sure something not None is returned if the SQL command is successful.

3) Tests are also much more complicated.

Also, is anyone interested in reviewing my other PR ([#621](https://github.com/biocore/american-gut-web/pull/621)) that updates the database? Thank you!

Best,
Qiyun

knimin/lib/data_access.py
```diff
@@ -2415,6 +2415,144 @@ def set_deposited_ebi(self):
...
+            If there is error editing the study's properties
+        """
+        with self._con.cursor() as cur:
+            cur.execute('BEGIN')
+            sql = """SELECT * FROM pm.study WHERE study_id = %s LIMIT 1"""
+            sql_args = [study_id]
+            if self._con.execute_fetchone(sql, sql_args) is None:
+                raise ValueError('Study ID %s does not exist.' % study_id)
+            if title:
+                sql = """SELECT study_id FROM pm.study WHERE study_id <> %s
```
antgonza: This could also be a method name_exists or exists_by_name

knimin/lib/data_access.py
```diff
@@ -2415,6 +2415,144 @@ def set_deposited_ebi(self):
...
+                if qry is not None:
+                    raise ValueError('Study title "%s" conflicts with another '
+                                     'study: %s.' % (title, qry[0]))
+            sql = """UPDATE pm.study SET title = %s, alias = %s, notes = %s
+                     WHERE study_id = %s RETURNING study_id"""
+            sql_args = [title or None, alias or None, notes or None, study_id]
+            if self._con.execute_fetchone(sql, sql_args) is None:
+                raise ValueError('Editing study %s failed.' % study_id)
+            cur.execute('COMMIT')
+        return True
```
antgonza: Same comment than before about returns.

knimin/lib/data_access.py
```diff
@@ -2415,6 +2415,144 @@ def set_deposited_ebi(self):
...
+            {title : str, alias : str, notes : str}
+
+        Raises
+        ------
+        ValueError
+            If there is error reading the study's properties
+        """
+        with self._con.cursor() as cur:
+            cur.execute('BEGIN')
+            sql = """SELECT title, alias, notes FROM pm.study
```
antgonza: This method could also take advantage of the exists method.

knimin/lib/data_access.py
```diff
@@ -2415,6 +2415,144 @@ def set_deposited_ebi(self):
...
+        bool
+            True if successful
+
+        Raises
+        ------
+        ValueError
+            If there is error deleting the study
+        """
+        with self._con.cursor() as cur:
+            cur.execute('BEGIN')
```
antgonza: Also take advantage of the exists method.

knimin/lib/data_access.py
antgonza: This method can be deleted now, right?

knimin/lib/data_access.py
```diff
@@ -2415,6 +2415,144 @@ def set_deposited_ebi(self):
                  WHERE barcode IN %s"""
         self._con.execute(sql, [barcodes])
 
+    def create_study(self, study_id, title=None, alias=None, notes=None):
+        """Creates a study
+
+        Parameters
+        ----------
+        study_id : int
+            Positive if it is a Qiita study ID, negative if others
```
qiyunzhu: This positive / negative system was suggested by @josenavas . @josenavas do you have any comments on this?

knimin/lib/data_access.py
```diff
@@ -2415,6 +2415,144 @@ def set_deposited_ebi(self):
...
+            True if successful
+
+        Raises
+        ------
+        ValueError
+            If there is error creating the study
+        """
+        with self._con.cursor() as cur:
+            cur.execute('BEGIN')
+            sql = """SELECT study_id, title FROM pm.study WHERE study_id = %s
```
qiyunzhu: This function is to add a new row in the table "study", i.e., a new study. Do that make sense?

knimin/lib/data_access.py
```diff
@@ -2415,6 +2415,144 @@ def set_deposited_ebi(self):
...
+            if qry is not None:
+                raise ValueError('Study ID or title conflicts with exisiting '
+                                 'study %s: %s.' % (qry[0], qry[1]))
+            sql = """INSERT INTO pm.study (study_id, title, alias, notes)
+                     VALUES (%s, %s, %s, %s) RETURNING study_id"""
+            sql_args = [study_id, title or None, alias or None, notes or None]
+            if self._con.execute_fetchone(sql, sql_args) is None:
+                raise ValueError('Creation of study %s failed.' % study_id)
+            cur.execute('COMMIT')
+        return True
```
qiyunzhu: I see! I am thinking too. Can a Python function not return anything? If so, how can I unit-test it?
I am thinking about having those functions to return the ID of the rows affected. For a new created object with a serial ID, it also returns the ID. Although I cannot predict what number will this be, I can do `self.assertGreater(obs, 0)` to test it. What do you think?

knimin/lib/data_access.py
```diff
@@ -2415,6 +2415,144 @@ def set_deposited_ebi(self):
...
+            True if successful
+
+        Raises
+        ------
+        ValueError
+            If there is error editing the study's properties
+        """
+        with self._con.cursor() as cur:
+            cur.execute('BEGIN')
+            sql = """SELECT * FROM pm.study WHERE study_id = %s LIMIT 1"""
```
qiyunzhu: You are right about using an **exists** method. I agree with you and hope that our codes would be generic, short and reusable. @josenavas and @wasade what do you think?
In these particular cases, however, there may be a trivial obstacle to using a generic `name_exists` function. The error messages I tried to raise include some information from the returned existing rows, and they differ in pattern. Therefore a simple name_exists function may not solve the problem.

knimin/lib/data_access.py
qiyunzhu: See above.

knimin/lib/data_access.py
qiyunzhu: See above.

knimin/lib/data_access.py
qiyunzhu: See above.

knimin/lib/data_access.py
antgonza: That's exactly my point: you are adding a new row (INSERT INTO pm.study) to a table that already has a row with those values (SELECT study_id, title FROM pm.study) ... why do that vs. update?

knimin/lib/data_access.py
antgonza: yes, no return is fine. We normally test that the new study is actually in the DB and that the id is the next incremental before the insertion.

knimin/lib/data_access.py
antgonza: name_exists: why can't you use the same error?

knimin/lib/data_access.py
qiyunzhu: Got it! Will edit function to make them return ID or no return. For ID, I read about Postgres, and had the impression that we can not predict the next number in an incremental sequence without actually creating it. What do you think?

knimin/lib/data_access.py
qiyunzhu: I think INSERT adds a new row, and UPDATE updates an existing row. Am I getting it right?

knimin/lib/data_access.py
antgonza: Retrieving the next incremental for general code is not recommended but we do this for testing code all the time, you can see an example [here](https://github.com/biocore/qiita/blob/master/qiita_db/test/test_util.py#L151).

knimin/lib/data_access.py
qiyunzhu: That's because, for example:
**create_study** wants to check either ID or name exists, and to tell user the conflicting record's name and ID.
**edit_study** wants to check if name exists but its ID is different, and to tell user you are reusing a duplicate name.
**read_study** and **delete_study** only check if name exists.
In the future these commands will also check other relevant tables, e.g., if some samples have been assigned to a study the user wants to delete.
Given these complicated situations, I didn't come up with a generic solution. Actually I would always prefer generic ones if there is any, even the performance is not optimal. Here seems to be a different case.

knimin/lib/data_access.py
qiyunzhu: Thanks! Let me read it. Will also read the implementation of `exists`.

knimin/lib/data_access.py
antgonza: You know something, I just realized that I'm reading this wrong, the trick here is: `if qry is not None:`.

qiyunzhu: Got it! So I don't have to modify anything here, do I?

knimin/lib/data_access.py
```diff
@@ -2415,6 +2415,126 @@ def set_deposited_ebi(self):
...
+        alias : str
+        notes : str
+
+        Raises
+        ------
+        ValueError
+            If there is error creating the study
+        """
+        with self._con.cursor() as cur:
+            cur.execute('BEGIN')
```
wasade: I believe this will establish the transaction. @josenavas, I do think you're the resident expert :) any concerns with isolation level or implicit behavior from executions which get run inside (e.g., the calls to `execute_fetchone`)?

knimin/lib/data_access.py
```diff
@@ -2415,6 +2415,126 @@ def set_deposited_ebi(self):
...
+        dict
+            {title : str, alias : str, notes : str}
+
+        Raises
+        ------
+        ValueError
+            If there is error reading the study's properties
+        """
+        with self._con.cursor() as cur:
+            cur.execute('BEGIN')
```
wasade: i dont believe this needs to be implemented within an explicit transaction as it is a single query. Additionally, it is not modifying the database

knimin/lib/data_access.py
josenavas: I don't think this is working as expected. `execute_fetchone` explicitly executes a commit, which means that those calls are either (1) executed in a different "transaction" since it is using a different cursor or (2) everything is committed as before.

I think the best path moving forward is actually copying the transaction object in Qiita to here (probably under a different module to avoid conflicts) and use it here, rather than implementing different transactions here, which is just gonna make the code less compliant with the other projects in the lab.

knimin/lib/data_access.py
qiyunzhu: @wasade Thank you! So operations that do not modify the database do not have to be included in a transaction. Am I getting it right?

knimin/lib/data_access.py
qiyunzhu: I guess I understand your rationale. @wasade and @josenavas please make a decision which way to go. I am not capable of implementing the Qiita transaction object. So can anyone do it if we want to go this way?

antgonza:
👍

qiyunzhu:
Antonio: Thank you!! Anyone else is interested in looking at the code? -
Qiyun

wasade:
I'm :+1:, @josenavas, you're the lingering reviewer!

Issue #89, opened 2016-08-23
Handlers missing tests


wasade:
- [x] access_control.py (@wasade, #86)
- [ ] access_decorators.py (@wasade, #88)
- [ ] ag_add_barcode_kit.py
- [ ] ag_consent_check.py
- [ ] ag_edit_barcode.py
- [ ] ag_edit_kit.py
- [ ] ag_edit_participant.py
- [ ] ag_get_participant_names.py 
- [ ] ag_new_barcode.py
- [ ] ag_new_kit.py
- [ ] ag_pulldown.py
- [ ] ag_results_ready.py
- [ ] ag_search.py
- [ ] ag_stats.py
- [ ] ag_third_party.py
- [ ] ag_update_geocode.py
- [ ] auth_handlers.py (@sjanssen2)
- [ ] barcode_util.py
- [ ] base.py (may not need to be directly tested)
- [ ] logged_in_index.py (@ElDeveloper)
- [ ] projects_summary.py

sjanssen2:
I tried to write tests for ag_edit_participant.py but I failed to open the according website on my browser, even after creating an admin user. Is this class unused?

wasade:
It's still a working page from what I can see. You can access it via search, and edit. To go direct, you need to specify an email as an argument

mortonjt:
I noticed that ag_stats.py is already tested [here](https://github.com/biocore/labadmin/blob/master/knimin/tests/test_ag_stats.py) and ag_third_party.py is tested [here](https://github.com/biocore/labadmin/blob/master/knimin/tests/test_third_party.py).

Are these tests considered inadequate, or should they be taken off of this list

wasade:
From a quick glance, there seems to be a bunch of stuff there. What are you
thoughts?

mortonjt:
From what I can tell, it looks like these two modules are already tested ...

wasade:
kk

sjanssen2:
I redesigned tests for ag_stats.py because it was testing against hard coded numbers. Now it compares to stats obtained from database queries. It is part of my PR #111 , which fails otherwise, since it makes unrevertable changes to the DB.

sjanssen2:
I'd like to know which lines of the code are not covered by my tests. I think this is normally possible with the coveralls page, but for this project it says: "The owner of this repo needs to re-authorize with github; their OAuth credentials are no longer valid so the file cannot be pulled from the github API.". Can we change this?

wasade:
[...]

PR #101, opened 2016-09-16
Unittest projects_summary.py

sjanssen2:


knimin/tests/test_auth_handlers.py
```diff
@@ -0,0 +1,66 @@
...
+        port = self.get_http_port()
+        self.assertEqual(response.effective_url, 'http://localhost:%d/' % port)
+
+    def test_post(self):
+        # check if unkown users are recognized
+        response = self.post('/auth/login/',
+                             {'email': 'idontexist',
+                              'password': 'password',
+                              })
+        self.assertIn('>Unknown user<', response.body)
```
wasade: if this does not return a non 200 status code, should it?

knimin/tests/test_auth_handlers.py
```diff
@@ -0,0 +1,66 @@
...
+
+class TestAuthLogoutHandler(TestHandlerBase):
+    def test_get(self):
+        response = self.get('/auth/login/')
+        self.assertEqual(response.code, 200)
+        # make sure redirect happened properly
+        port = self.get_http_port()
+        self.assertEqual(response.effective_url, 'http://localhost:%d/' % port)
+
+        # TODO: how can I check that the cookie has been cleared? Or don't I
```
wasade: Could just do it implicitly by trying to access a page which requires auth

knimin/tests/test_projects_summary.py
```diff
@@ -0,0 +1,28 @@
+from unittest import main
+
+from knimin import db
+from knimin.tests.tornado_test_base import TestHandlerBase
+
+
+class TestProjectsSummaryHandler(TestHandlerBase):
+    def test_get_not_authed(self):
+        response = self.get('/projects/summary/')
+        # print(response)
```
wasade: remove please

knimin/tests/test_auth_handlers.py
sjanssen2: That's what I would expect, but it doesn't. The return code is always "200", regardless of the success of login. I think we should change this later on to a more speaking return code, but for the current unit tests I have to work with whatever is implemented.

knimin/tests/test_auth_handlers.py
sjanssen2: That was my initial idea, but isn't there a difference between a logged in user and a stored cookie?

knimin/tests/test_projects_summary.py
sjanssen2: up, will do.

sjanssen2:
How can I update the text for issue #89 ? Could it be, that I don't have the required permissions? I don't see an edit option anywhere.

knimin/tests/test_auth_handlers.py
wasade: I'd be in favor of a modification to the handler in this respect, what do you think? Or, we can defer to a future issue, but I suspect change is light

wasade:
Should be fixed. Let me know if that didn't work to allow edit permissions.
If it doesn't, I'll just edit

wasade:
Do we care about a cookie if the logged out user isn't able to access
authorized pages?

knimin/tests/test_auth_handlers.py
sjanssen2: I can't foresee which other parts might break due to this modification. Unittest are exactly for this purpose, don't they? So I would prefer to first complete all unit tests and only then modify the code - but we should raise a new issue here.

wasade:
okay, sounds good

knimin/tests/test_auth_handlers.py
sjanssen2: opened issue #102

sjanssen2:
stupid me. I added the auth_handler.py to this PR, but it is topic of PR #100  :-/

wasade:
Thanks @sjanssen2!!

PR #104, opened 2016-09-16
Unittest ag_get_participant_names

sjanssen2:


knimin/tests/test_ag_get_participant_names.py
```diff
@@ -0,0 +1,47 @@
...
+                         'http://localhost:%d/ag_participant_names/' % port)
+
+
+class TestAGNamesDLHandler(TestHandlerBase):
+    def test_post(self):
+        self.mock_login_admin()
+        response = self.post('/ag_participant_names/download/', {})
+        self.assertEqual(response.code, 200)
+        # is it sufficient to check that the response provides a zip-file or
+        # should I also unpack the content and compare it to a database query?
```
wasade: I think this is fine

wasade: actually, that would be useful if easy. however, if the underlying logic is already tested separately then it is redundant

sjanssen2: OK then. I don't check here and defer that test to another point in the code.

PR #108, opened 2016-09-19
Unittest ag new barcode

sjanssen2:


knimin/tests/test_ag_new_barcode.py
```diff
@@ -0,0 +1,131 @@
...
+        response = self.get('/ag_new_barcode/download/')
+        self.assertEqual(response.code, 405)  # Method Not Allowed
+
+    def test_post(self):
+        self.mock_login_admin()
+        response = self.post('/ag_new_barcode/download/',
+                             {'barcodes': "1111,222,33,4"})
+        self.assertEqual(response.code, 200)
+        # is it sufficient to check that the response provides a pdf-file or
+        # should I somehow check the contents of the PDF?
```
mortonjt: I think it is ok to just check to see if the pdf exists and is not empty.

wasade: agree, on the assumption the contained functionality is reasonably tested. checking the pdf would be a pain and i dont think worth the effort

wasade:
seems good, @mortonjt, you :+1:?

PR #109, opened 2016-09-19
Adding testing instructions

mortonjt:


README.rst
```diff
@@ -36,3 +36,18 @@ Initial default test login credentials are:
 
 **Password:** password 
 
+Testing
+-------
+
+To run the webserver locally, first run the following commands to insert a user with administrative access::
+
```
sjanssen2: Can you also add the SQL commands to select the right database.

README.rst
```diff
@@ -36,3 +36,18 @@ Initial default test login credentials are:
...
+    INSERT INTO ag.labadmin_users (email, password) VALUES ('master', '$2a$10$2.6Y9HmBqUFmSvKCjWmBte70WF.zd3h4VqbhLMQK1xP67Aj3rei86');
+    
+    INSERT INTO ag.labadmin_users_access (access_id, email) VALUES (7, 'master');
+    
+Then you should be able to login with the following credentials
+
+**User:** master
+
+**Password:** password 
+
```
sjanssen2: Maybe add a note about the test behaviour, i.e. the access of the "test" user is re-set after each unit test.

README.rst
```diff
@@ -36,3 +36,21 @@ Initial default test login credentials are:
...
+Note that whenever a unittest is run, the user `test` is granted admin rights, which is reset once the unittest is completed.
```
wasade: I believe this is only the case if `mock_admin_login` is used

mortonjt: Right

josenavas:
How is the change on the README.md changing the coverage to +6.1%?

mortonjt:
It was a very well tested readme :D

README.rst
mortonjt: Done!

README.rst
mortonjt: Done!

mortonjt:
All comments have been addressed.  Anything else left here?

README.rst
```diff
@@ -36,3 +36,21 @@ Initial default test login credentials are:
...
+    
+    INSERT INTO ag.labadmin_users_access (access_id, email) VALUES (7, 'master');
+    
+Then you should be able to login with the following credentials
+
+**User:** master
+
+**Password:** password 
+
+Note that whenever a unittest is run, the user `test` is granted admin rights, which is reset once the unittest is completed if `mock_admin_login` is called.
```
wasade: This is still inaccurate. The user `test` is not implicitly granted admin rights

mortonjt: Can you clarify exactly how this should be corrected?

wasade: Delete it? The statement is inaccurate

sjanssen2: To me, Jamie's description seems to be accurate, since tearDown() alters the 'test' user access level:
`    def tearDown(self):
        BaseHandler.get_current_user = self.orig_func
        # Remove all access privileges user may hve been given by a test
        db.alter_access_levels('test', [])
        super(TestHandlerBase, self).tearDown()
`
Which means that a manual change of the access rights for this user are overwritten by any unit test.

sjanssen2: oh. I see. The first part is not precise enough. 'test' only get admin rights if `mock_login_admin` or its manual equivalent `db.alter_access_levels('test', [7])` is used, but not per default for every test in general.

sjanssen2: Let us rephrase the sentence like that:
"After executing our existing unit test suite, access level for the user 'test' will be reset to '', i.e. he won't be able to see most of the main menu items. Thus, adding a second user 'master' with admin privileges is quite useful."

mortonjt: Done.  Thanks @sjanssen2 !  @wasade do you have any thoughts on this current phrasing?

sjanssen2:
@josenavas @wasade would be cool to come to a decision rather we want to merge or not.

PR #110, opened 2016-09-19
adding test for ag_update_geocode.py

qiyunzhu:


mortonjt:
👍  @qiyunzhu is this ready to merge?

qiyunzhu:
@mortonjt I think so. Tested and working. I hope I didn't miss anything. What do you think?

mortonjt:
I love the squash and merge feature :)

knimin/tests/test_update_geocode.py
```diff
@@ -0,0 +1,27 @@
...
+            '/login/?next=%2Fag_update_geocode%2F'))
+
+    def test_get(self):
+        self.mock_login()
+        response = self.get('/ag_update_geocode/')
+        self.assertEqual(response.code, 200)
+
+        # check that correct information is printed on HTML page.
+        result = db.getGeocodeStats()
+        for i in range(len(result)):
```
josenavas: This is not pythonic. The pythonic way of doing this is:
```python
for x in result:
    self.assertIn('<td>%s</td><td>%s</td>' % (x[0], x[1]), response.body)
```

mortonjt:
It's not a big deal.  Want to revert?

josenavas:
nah, no need to revert since it is on the tests, but just thought to point it out since that way of doing the for loops in "real" code can have an impact on performance.

Issue #102, opened 2016-09-16
knimin/handlers/auth_handlers.py should return != 200 for wrong pwd or user


sjanssen2:
Logging in with wrong credentials is not possible and the user gets notified by a warning on the website. However, the return code in those cases should be different from 200, which indicates that everything is OK. 
Thus, unit tests would be much more clean.

sjanssen2:
I read that this is the default behaviour of tornado. For get() methods without valid credentials it will return a 200 code. Only for post() methods it is 405. I think we can't and won't really change this. Thus, I am closing this issue

PR #111, opened 2016-09-19
Unittest ag new kit

sjanssen2:


knimin/tests/test_ag_new_kit.py
```diff
@@ -0,0 +1,129 @@
...
+        response = self.post('/ag_new_kit/',
+                             {'tag': 'toolongtag',
+                              'projects': ['PROJECT2', 'PROJECT5'],
+                              'swabs': swabs,
+                              'kits': kits,
+                              })
+        self.assertEqual(response.code, 500)
+        self.assertIn("Tag must be 4 or less characters", response.body)
+
+        # check project existens
```
wasade: existens?

knimin/tests/test_ag_new_kit.py
```diff
@@ -0,0 +1,129 @@
...
+        self.assertEqual(response.code, 400)
+
+        # too long tag
+        response = self.post('/ag_new_kit/',
+                             {'tag': 'toolongtag',
+                              'projects': ['PROJECT2', 'PROJECT5'],
+                              'swabs': swabs,
+                              'kits': kits,
+                              })
+        self.assertEqual(response.code, 500)
```
wasade: for this and below, a `500` means something went wrong that is unexpected. Do you think we should revise the handler to gracefully handle this and the below?

knimin/tests/test_ag_new_kit.py
```diff
@@ -0,0 +1,129 @@
...
+
+        # check for empty swabs list
+        response = self.post('/ag_new_kit/',
+                             {'tag': tag,
+                              'projects': ['PROJECT2', 'PROJECT5'],
+                              'swabs': [],
+                              'kits': kits,
+                              })
+        self.assertEqual(response.code, 500)
+        self.assertIn("SET assigned_on = NOW() WHERE barcode IN ()",
```
wasade: I don't think we ever should present the user SQL...?

knimin/tests/test_ag_new_kit.py
sjanssen2: I would prefer to first complete tests for the existing - maybe crappy - code base and only start improving it afterwards, as we already decided for issue #102

knimin/tests/test_ag_new_kit.py
sjanssen2: Agree. But we do at the moment, because we are to lazy to capture this situation in the handler or the python connector to SQL :-/ Add this to our todo list.

knimin/tests/test_ag_new_kit.py
sjanssen2: Opened issue #113

knimin/tests/test_ag_new_kit.py
sjanssen2: see issue #113

wasade:
:+1:

PR #112, opened 2016-09-19
TST: Adding tests for ag_search

mortonjt:


knimin/tests/test_ag_search.py
```diff
@@ -0,0 +1,46 @@
...
+        if details['withdrawn'] is None:
+            payload['withdrawn'] = 'N'
+        else:
+            payload['withdrawn'] = details['withdrawn']
+
+        payload['search_term'] = 'test'
+        response = self.post('/ag_search/', payload)
+        self.assertIn('TESTDUDE', response.body)
+        self.assertIn('TESTOTHER', response.body)
+        self.assertIn('United Kingdom', response.body)
```
wasade: 000004216 has a UK country? That's weird, because that's one of my barcodes

mortonjt: Hmm.  I'm assuming that it is a test database?

wasade: it's because the `search_term` is so broad. why are all the `getAGBarcodeDetails` bits passed in, it doesn't looks necessary? I think `payload` only needs to contain `search_term`

mortonjt: Ok, I can remove it.

mortonjt: Why are the getAGBarcodeDetails passed in?

That I don't know - the post command explicitly required this ...
Maybe we should refactor the post command?

wasade: It looks like it only requires `search_term`? https://github.com/biocore/labadmin/blob/master/knimin/handlers/ag_search.py#L18

sjanssen2: I don't understand what is going on with the "payload". To me, it seams to be unnecessary - but I might be wrong here. I submitted alternative unit tests with PR #123

mortonjt:
Closing this in favor for #123

knimin/tests/test_ag_search.py
mortonjt: Hmm - I'll need to ask about this later.  I was getting a whole bunch of errors when I didn't specify a payload.

PR #116, opened 2016-09-20
TST: tests for ag_results_ready

wasade:


mortonjt:
👍

sjanssen2:
👍 from my side. However, coverage is not 100%, but we can do that in another PR.

PR #121, opened 2016-09-22
DO NOT MERGE

sjanssen2:


sjanssen2:
@wasade @josenavas I don't understand how the coverage is computed. How can the file knimin/lib/string_converter.py have a coverage of 38.89% after I commented all tests out in the according knimin/lib/tests/test_string_converter.py ?

sjanssen2:
I also don't understand why coverage is not 100%, because it looks to me that all functions are triggered more than once.

wasade:
Indirect execution, eg https://github.com/biocore/labadmin/blob/master/knimin/lib/data_access.py#L1048

sjanssen2:
Oh I see. I commented out all those lines, because I don't understand why the coverage is not 100% for this module. Might it be the case, that the according unit tests in knimin/lib/tests/test_string_converter.py are never executed? If so, is that a bug in some config file?

wasade:
not sure of a config file which would effect this, only one i can think that might is [here](https://github.com/biocore/labadmin/blob/master/.coveragerc). how does coverage look when you run locally? if you run the tests directly, do they actually run?

sjanssen2:
I found the problem, the executable bit for this test file was set, thus nose did not want to run these tests :-)

wasade:
i love and hate nose...

wasade:
good find btw :)

PR #122, opened 2016-09-22
Unittest ag results ready

sjanssen2:
test situations where there 
a) are no barcodes in the dir
b) is no config dir


knimin/tests/test_ag_results_ready.py
```diff
@@ -25,13 +27,29 @@ def test_post(self):
...
             cur.execute("""SELECT results_ready
                            FROM ag.ag_kit_barcodes
                            WHERE barcode='000023299'""")
             result = cur.fetchone()[0]
             self.assertEqual(result, 'Y')
 
+        # test that error is raised if pdf directory does not exist
+        db.config.base_data_dir = 'phantasy_path'
+        response = self.post('/update_ready/', {})
+        self.assertEqual(response.code, 500)
```
wasade: ideally we never 500, although it does seem appropriate in this situation. what do you think about presenting the user with a message about a bad configuration?

sjanssen2: I opend issue #126 and added TODO comments at the handler and test code.

PR #123, opened 2016-09-22
Unittest ag search

sjanssen2:
I might be wrong here, but I don't understand what @mortonjt is doing in PR #112 with all the payload stuff. Sorry Jamie. Here is an alternative unit test for the same handler.


mortonjt:
That's ok - I don't know what I'm doing either :/  Closing #112 

Can you add a test for searching for general terms?  You aren't limited to just barcodes and kit ids.
Once that is done, I think this is ready to go.

sjanssen2:
We can search for arbitrary terms. However, from reading the code, I got the impression that only objects are reported which are "samples", "kits" or "users". I'd be happy to see a search term example where something else than that is printed to the HTML side.

mortonjt:
OK.  In that case, this should be good to go.  @wasade what do you think?

knimin/tests/test_ag_search.py
```diff
@@ -0,0 +1,70 @@
...
+    def test_post_user(self):
+        self.mock_login_admin()
+
+        # search participant info
+        search_term = 'qdkwzpkuci@yrvxr.com'
+        response = self.post('/ag_search/', {'search_term': search_term})
+        self.assertEqual(response.code, 200)
+        for login in db.search_participant_info(search_term):
+            for l in db.get_login_info(login):
+                for field in l.keys():
```
wasade: minor, but the call to `.keys()` is not necessary. same on the other uses below

wasade:
one comment, :+1: otherwise

PR #124, opened 2016-09-23
Unittest ag update geocode

sjanssen2:
added test for post method + renaming to match handler file name


mortonjt:
Something looks funny with the auth test.  Do you expect to get a 200 message if the auth fails?

sjanssen2:
@mortonjt I agree, see Issue #102

PR #128, opened 2016-09-23
found an issue with user access level + added testing an edge case

sjanssen2:


knimin/handlers/access_control.py
```diff
@@ -33,6 +33,8 @@ def post(self):
 
         all_levels = db.get_access_levels()
         user_levels = db.get_access_levels_user(user)
+        # TODO: I think the condition must check the access level of the logged
```
wasade: is this check even necessary given the handler decorator?

sjanssen2: As far as I know, `@authenticated` assures that a valid user has been logged in, independent of its role in our system. Thus, a user with only "scan barcodes" role might be able to access this web site, but he should not have the right to change other users access rights. 
Thus, it is necessary to have this check and we must correct it here. Otherwise, an admin can only edit other admin (or himself) rights.

wasade: I meant the class level decorator which defines the roles required iirc

sjanssen2: oh. I never noticed that decorator. Looks like you are right and this if statement is not necessary at all.

PR #133, opened 2016-09-26
Unittest squash barcodes

sjanssen2:
To check all conditions in the build_barcodes_pdf method, I had to dynamically replace the get_image method. Thus, I needed to change how those methods are imported.

### DIFF
--- a/knimin/lib/tests/test_squash_barcodes.py
+++ b/knimin/lib/tests/test_squash_barcodes.py
@@ -15,9 +15,9 @@ class SquashBarcodesTests(TestCase):
 
         # replace get_image method to change behaviour
         def mock_get_image1(barcodes):
-            font = join(dirname(realpath(knimin.lib.__file__)), 'FreeSans.ttf')
+            font = join(dirname(realpath(__file__)), '../', 'FreeSans.ttf')
             for b in barcodes:
-                yield code128_image(b, height=150, width=300, font=font
+                yield code128_image(b, height=150, width=300, font=font,
                                     thickness=2, show_text=True,
                                     quiet_zone=False)
         m.get_image = mock_get_image1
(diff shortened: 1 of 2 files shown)


knimin/lib/tests/test_squash_barcodes.py
```diff
@@ -1,28 +1,68 @@
...
+        # reset function
+        m.get_image = old_get_image
+
+    def test_tmpdir(self):
+        # test that a temporary directory cannot collide with an existing one
+        random.seed(7)
+        fake_tmp_dir = 'tmp_IDQBN'
+        if not path.exists(fake_tmp_dir):
+            mkdir(fake_tmp_dir)
+        self.assertTrue(m.build_barcodes_pdf(['000000011']))
```
wasade: it doesn't look like this test has anything to do with the `tmp_dir`? I'm a bit confused

On a related note, the `tempfile` provides excellent functionality for generating temp files or directories with mechanisms to ensure they're cleaned up if desired

sjanssen2: The `build_barcodes_pdf` method contains the somewhat strange loop `    while path.exists(tmpdir):
        tmpdir = "tmp_%s" % ''.join([choice("ABCDEFGHIJKLMNOPQRSTUVWXYZ")
                                     for i in range(5)])
    mkdir(tmpdir)
` to make sure creating a new, non existing tmp dir. I want to check if this look functions as expected by first creating a colliding dir, such that the loop must be executed at least once.
I agree, it is very confusing.

wasade: "somewhat strange" is polite. it would be great to just leverage python's `tempfile` module

wasade:
Appears there are merge conflicts

sjanssen2:
@wasade : conflicts resolved!

PR #134, opened 2016-09-26
Unittest ag edit barcode

sjanssen2:


knimin/tests/test_ag_edit_barcode.py
```diff
@@ -0,0 +1,108 @@
...
+from tornado.escape import url_escape
+
+from knimin.tests.tornado_test_base import TestHandlerBase
+from knimin import db
+
+
+class TestAGEditBarcodeHandler(TestHandlerBase):
+    def test_get_not_authed(self):
+        response = self.get('/ag_edit_barcode/')
+        self.assertEqual(response.code, 200)
```
wasade: I thought there was an issue about this but I'm not seeing it right now -- should unauthorized attempts result in a `200` or `403`?

knimin/tests/test_ag_edit_barcode.py
```diff
@@ -0,0 +1,108 @@
...
+                         'http://localhost:%d/login/?next=%s' %
+                         (port, url_escape('/ag_edit_barcode/')))
+
+    def test_get(self):
+        self.mock_login_admin()
+
+        # check that error is raised for unknown barcode
+        response = self.get('/ag_edit_barcode/', {'barcode': 'unknown'})
+        self.assertEqual(response.code, 500)
+        self.assertRaises(KeyError)
```
wasade: Does this actually work as expected? If the call to `self.get` raises a `KeyError`, the exception will happen at that point in the stack. Generally, the form used with `assertRaises` is:

```python
with self.assertRaises(SomeSpecificException):
    some_block_of_code_which_excepts
```

The context manager than catches the exception and tests that it is the expected exception

knimin/tests/test_ag_edit_barcode.py
```diff
@@ -0,0 +1,108 @@
...
+                              'site_sampled': 'Stool',
+                              'sample_date': '2015-01-17',
+                              'sample_time': '08:00:00',
+                              'participant_name': "REMOVED-0",
+                              'notes': "REMOVED",
+                              'refunded': 'N',
+                              'withdrawn': 'N',
+                              'environment_sampled': None,
+                              })
+        self.assertEqual(response.code, 200)
```
wasade: Might want a TODO or issue here on status code

knimin/tests/test_ag_edit_barcode.py
sjanssen2: documentation of tornado specifies that a 200 is returned for an unauthed user for a get method, but a 403 for an unauthed user for a post method. Thus, it is working as specified - but maybe not as expected. But I guess we should not strive to change this basic behaviour of tornado.

knimin/tests/test_ag_edit_barcode.py
sjanssen2: surprisingly it works :-) Could it be that first a page is returned with code 500 and after that an exception is raised by tornado? Anyways, I don't see a reason to change the code here, do you agree?

knimin/tests/test_ag_edit_barcode.py
sjanssen2: added a todo and raised an issue to keep track of it

knimin/tests/test_ag_edit_barcode.py
wasade: okay, agre

wasade:
There are merge conflicts.

@sjanssen2, I'm okay with merge once the `self.assertRaises` is changed. I verified independently that the above code with a `self.assertRaises(KeyError)` in isolation doesn't actually do anything.

PR #135, opened 2016-09-26
Unittest configuration

sjanssen2:
adding one more test to see if exception is raised if not all sections of the config present in file.


knimin/lib/tests/test_configuration.py
```diff
@@ -22,6 +22,14 @@ def test_init(self):
         with self.assertRaises(IOError):
             KniminConfig('does not exist')
 
+        # test that expection is raised if not all sections are specified
+        self.config = tempfile.NamedTemporaryFile()
```
wasade: Would it be possible to set a local variable instead of `self.config`? While `setUp` should ensure this variable is reset for each test, the use of `self.` implies the modified state is intended for consumption outside of the local scope

sjanssen2: Unfortunately, it cannot be just a locale variable, because the method itself is directly using self.config and not an argument. However, since I test here the __init__ of the whole config object and trigger an exception this object should die directly after that test and should not be available anywhere else. Or am I wrong here?
I think, reading an incomplete config file is such a fundamental error, that either the system should stop right here (which it does) or be refactored!
Thus, I think we are fine with this dangerous move here, but I totally agree that we should avoid hacking an object like that in general.

wasade: are you _sure_ the logic behaves differently if `self` is remove? ie:

```python
config = tempfile.NamedTemporaryFile()
config.write(test_config[:100])
config.seek(0)
config_fp = config.name
with self.assertRaises(ValueError):
    KniminConfig(config_fp)
```

sjanssen2: I checked. You are right. The tests also works without self. I changed the according code block as you suggested. Thanks for enlighten me here!

PR #136, opened 2016-09-26
added one more test

sjanssen2:


knimin/tests/test_ag_new_kit.py
```diff
@@ -124,5 +124,17 @@ def test_post(self):
...
+        # what if tag is None
+        response = self.post('/ag_new_kit/',
+                             {'tag': '',
+                              'projects': ['PROJECT2', 'PROJECT5'],
+                              'swabs': swabs,
+                              'kits': kits,
+                              })
+        self.assertEqual(response.code, 200)
+        kitinfo = loads(response.body)
+        self.assertNotIn('_', kitinfo['kitinfo'][0][0])
```
wasade: this seems rather non-specific?

sjanssen2: true ... at first. I want to test that there is no prefix (= "tag") set. The kit ID is composed of <tag>_<random> if tag == None than we don't get the _ as well. Thus I think it is the only way to check for the absence of a tag. Or do you see another way?

PR #140, opened 2016-09-29
Plate mapper add Qiita TRN

qiyunzhu:


qiyunzhu:
Hi all, thanks to Jose's kind instruction, I have successfully implemented Qiita's transaction module into labadmin. Now the transaction function works as expected. Therefore I rewrote the functions relevant to "study". Following Jose's suggestion, I simplified the Python code, moving repeated statements into separate functions. One of them looks awkward to me, though. Please let me know if you are all comfortable with the new version. Thank you! - Qiyun

PR #141, opened 2016-09-29
Plate mapper kiwi checkpoint 02

qiyunzhu:


qiyunzhu:
Hi all, thanks to Jose's kind instruction, I have successfully implemented Qiita's transaction module into labadmin. Now the transaction function works as expected. Therefore I rewrote the functions relevant to "study". Following Jose's suggestion, I simplified the Python code, moving repeated statements into separate functions. One of them looks awkward to me, though. Please let me know if you are all comfortable with the new version. Thank you! - Qiyun

knimin/lib/data_access.py
```diff
@@ -2415,6 +2405,58 @@ def set_deposited_ebi(self):
                  WHERE barcode IN %s"""
         self._con.execute(sql, [barcodes])
 
+    def _study_exists(self, study_id):
+        """Cofirms that a study ID exists
```
antgonza: Confirms

knimin/lib/data_access.py
```diff
@@ -2415,6 +2405,58 @@ def set_deposited_ebi(self):
...
+        Parameters
+        ----------
+        study_id : int
+
+        Raises
+        ------
+        ValueError
+            If the study ID does not exist
+        """
+        sql = """SELECT EXISTS (SELECT 1 FROM pm.study WHERE study_id = %s)"""
```
antgonza: Wondering, why not SELECT study_id?

knimin/lib/data_access.py
qiyunzhu: Good catch!

knimin/lib/data_access.py
antgonza: Agree on that statement but my question is why `SELECT EXISTS (SELECT 1 FROM pm.study WHERE study_id = %s)` vs. `SELECT EXISTS (SELECT study_id FROM pm.study WHERE study_id = %s)`?

qiyunzhu: I found this trick from [here](http://stackoverflow.com/questions/7471625/fastest-check-if-row-exists-in-postgresql). I am not aware whether this actually improves performance. @wasade @josenavas any comments?

qiyunzhu:
Dear all,

I appended functionalities regarding "samples" into this PR as well. Happy reviewing!

Note: for all "samples" functions, I have them to work on multiple samples all at once. My rationale is: it will be a very common situation that multiple samples are affected in one operation, and it is important to ensure the atomicity of this entire operation. For example, the user wants to create 96 samples. After 50 samples have already been created, the 51 one has an invalid sample ID, and the whole operation is aborted. Then we need to make sure the first 50 samples are cleared out.

Note: so far "studies" and "samples" are not linked up together. I will add these functionalities in the next PR.

Best,
Qiyun

josenavas:
@qiyunzhu can you pull from upstream plate-mapper-kiwi and update this PR, thanks!

qiyunzhu:
I fixed it. Now here is the new PR:
https://github.com/biocore/labadmin/pull/142

josenavas:
Thanks! For future reference, note that you don't need to close the PR, if you pull, fix the merge conflicts and push again the PR will update :-)

qiyunzhu:
Sorry! That was because I wasn't sure how to resolve conflict properly. So
I deleted the entire branch and created a new one :P

PR #142, opened 2016-09-30
plate mapper 02 conflict resolved

qiyunzhu:


knimin/lib/data_access.py
```diff
@@ -2404,6 +2405,58 @@ def set_deposited_ebi(self):
...
+            Search for these values to make sure they don't already exist
+        study_id : int
+            Skip this ID in searching
+
+        Raises
+        ------
+        ValueError
+            If neither value is given, or either value already exists
+        """
+        cols = ((qiita_study_id, 'qiita_study_id', 'Qiita study ID'),
```
wasade: minor and non-blocking, but a different take on this:

```python
if qiita_study_id is None and title is None:
    # ... raise value error

for id_, column in zip((qiita_study_id, 'qiita_study_id'), (title, 'title')):
    # do stuff
```

...I don't think the description of the column adds much in my opinion

knimin/lib/data_access.py
```diff
@@ -2530,19 +2554,168 @@ def delete_study(self, study_id):
...
+            for sample in samples:
+                sample_id = sample['id']
+                if self._sample_exists(sample_id):
+                    raise ValueError('Sample ID %s already exists.'
+                                     % sample_id)
+                barcode = sample.get('barcode') or None
+                if barcode:
+                    self._barcode_exists(barcode)
+                is_blank = sample.get('is_blank') or False
+                notes = sample.get('notes') or None
```
wasade: `.get` defaults to `None`, ie `notes = sample.get('notes')` should be fine

knimin/lib/data_access.py
```diff
@@ -2530,19 +2554,168 @@ def delete_study(self, study_id):
...
+        with TRN:
+            for sample in samples:
+                sample_id = sample['id']
+                if self._sample_exists(sample_id):
+                    raise ValueError('Sample ID %s already exists.'
+                                     % sample_id)
+                barcode = sample.get('barcode') or None
+                if barcode:
+                    self._barcode_exists(barcode)
+                is_blank = sample.get('is_blank') or False
```
wasade: `is_blank = sample.get('is_blank', False)`

knimin/lib/data_access.py
```diff
@@ -2530,19 +2554,168 @@ def delete_study(self, study_id):
...
+        with TRN:
+            for sample in samples:
+                sample_id = sample['id']
+                if not self._sample_exists(sample_id):
+                    raise ValueError('Sample ID %s does not exist.'
+                                     % sample_id)
+                barcode = sample.get('barcode') or None
+                if barcode:
+                    self._barcode_exists(barcode)
+                is_blank = sample.get('is_blank') or False
```
wasade: same comment here

knimin/lib/data_access.py
```diff
@@ -2530,19 +2554,168 @@ def delete_study(self, study_id):
...
+            }
+
+        Raises
+        ------
+        ValueError
+            If there is error editing the samples' properties
+        """
+        with TRN:
+            for sample in samples:
+                sample_id = sample['id']
```
wasade: minor but the lines 2652 - 2660 are replicated and could be decomposed but i do not think that should inherently blcok

wasade:
or @antgonza

antgonza:
I think it looks fine. @qiyunzhu, could you address @wasade's comments, I think they add clarity to the code. Once those are solved, 👍

knimin/lib/data_access.py
qiyunzhu: Got it! Thanks!

knimin/lib/data_access.py
qiyunzhu: Got it!

knimin/lib/data_access.py
qiyunzhu: Thanks! I didn't know the 2nd trick before.

knimin/lib/data_access.py
qiyunzhu: I see that they are duplicated. I couldn't figure out an easy way to move them into a separate function and to reduce lines of codes. Originally I had everything (create/edit/delete) in one single function. @antgonza indicated (?) that it may be a better idea to create classes like "study", "sample", "plate". I like this idea. For now maybe let me continue with what I did so that we can save some time.

qiyunzhu:
Thank you! @wasade and @antgonza . For the TRN codes, actually @josenavas typed one example on my laptop, and I followed it. So I think he will be fine with them.

qiyunzhu:
@antgonza I didn't actually change the code when replying to Daniel's comments. After I made changes I noticed that you already merged the PR. So I created a new PR ([#143](https://github.com/biocore/labadmin/pull/143)).

I made some more edits to the function `_study_is_unique` than @wasade suggested. The current status may be the simplest and quite reusable. The word "qiita" does not even occur twice in the entire function.

Howevery the Travis test didn't pass. Here is the error message:

```
FAIL: test_get (knimin.tests.test_ag_pulldown.testUpdateEBIStatusHandler)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/travis/build/biocore/labadmin/knimin/tests/test_ag_pulldown.py", line 28, in test_get
    self.assertEqual(response.code, 200)
AssertionError: 599 != 200
-------------------- >> begin captured logging << --------------------
tornado.general: WARNING: tornado.autoreload started more than once in the same process
tornado.access: INFO: 200 GET /update_ebi/ (127.0.0.1) 20012.48ms
--------------------- >> end captured logging << ---------------------

```
I didn't edit this file at all. I am not sure why it happens. Do you have any idea? Thanks!

wasade:
That's a network timeout. Weird. Suggest restarting the tests, I think its
a fluke

antgonza:
Yes, that's what I did and the reason I merged. However, it might be a good idea to add some sleep(5) in the tests to make sure this doesn't happen.

qiyunzhu:
Thanks for the opinions. I am sending a new pull request to see if it
happens again. - Qiyun

PR #143, opened 2016-10-01
Plate mapper 02 edit after Daniel's comment

qiyunzhu:


qiyunzhu:
@antgonza I didn't actually change the code when replying to Daniel's comments. After I made changes I noticed that you already merged the PR. So I created a new PR.

I made some more edits to the function `_study_is_unique` than @wasade suggested. The current status may be the simplest and quite reusable. The word "qiita" does not even occur twice in the entire function.

antgonza:
Cool, thanks.

PR #144, opened 2016-10-03
Plate mapper 03 added sample plate and layout

qiyunzhu:
Here are some new functions that handle "sample_plate" and "sample_plate_layout". They work seamlessly. Deleting a sample plate will also delete its associated layout. I also edited the "delete_samples" function. It will raise error if the sample is found to be associated with a sample plate.

### DIFF
--- a/knimin/lib/data_access.py
+++ b/knimin/lib/data_access.py
@@ -2426,16 +2426,18 @@ class KniminAccess(object):
 
         Parameters
         ----------
-        qiita_study_id : int
-        title : str
-        skip_id : int
-            Skip this ID in searching
+        qiita_study_id : int, optional
+        title : str, optional
+            Either qiita_study_id or title must be given to identify a study
+        skip_id : int, optional
+            Skip this study ID in searching
+            In function create_study, this is not used
+            In function edit_study, this is the current study to be edited
 
         Raises
         ------
         ValueError
-            If neither value is specified, or
-            If either value already exists
+            If neither value is given, or either value already exists
         """
         tags = ['Qiita study ID', 'Title']
         cols = [tag.lower().replace(' ', '_') for tag in tags]
@@ -2473,10 +2475,11 @@ class KniminAccess(object):
 
         Parameters
         ----------
-        qiita_study_id : int
-        title : str
-        alias : str
-        notes : str
+        qiita_study_id : int, optional
+        title : str, optional
+            Either qiita_study_id or title must be specified
+        alias : str, optional
+        notes : str, optional
 
         Returns
         -------
@@ -2500,10 +2503,11 @@ class KniminAccess(object):
         ----------
         study_id : int
             ID of the study to edit
-        qiita_study_id : int
-        title : str
-        alias : str
-        notes : str
+        qiita_study_id : int, optional
+        title : str, optional
+            Either qiita_study_id or title must be specified
+        alias : str, optional
+        notes : str, optional
         """
         self._study_exists(study_id)
         self._study_is_unique(qiita_study_id, title, study_id)
@@ -2595,10 +2599,10 @@ class KniminAccess(object):
         ----------
         samples : list of dict
             {
-             id : str (mandatory),
-             is_blank : bool (optional, default: False),
-             barcode : str (optional),
-             notes : str (optional)
+             id : str,
+             is_blank : bool, optional, default: False,
+             barcode : str, optional,
+             notes : str, optional
             }
 
         Raises
@@ -2632,9 +2636,9 @@ class KniminAccess(object):
             {
              id : str,
                 ID of the sample to edit
-             is_blank : bool (optional, default: False),
-             barcode : str (optional),
-             notes : str (optional)
+             is_blank : bool, optional, default: False,
+             barcode : str, optional,
+             notes : str, optional
             }
 
         Raises
@@ -2746,7 +2750,7 @@ class KniminAccess(object):
         Parameters
         ----------
         name : str
-        skip_id : int
+        skip_id : int, optional
             Skip this ID in searching
 
         Raises
@@ -2768,7 +2772,7 @@ class KniminAccess(object):
 
         Parameters
         ----------
-        barcode : str
+        email : str
 
         Raises
         ------
@@ -2780,14 +2784,13 @@ class KniminAccess(object):
         if not self._con.execute_fetchone(sql, [email])[0]:
             raise ValueError('Email %s does not exist.' % email)
 
-    def _plate_type_name_to_id(self, name):
+    def _plate_type_name_to_id(self, name=None):
         """Gets plate type ID by name
 
         Parameters
         ----------
-        name : str
-            Name of the plate type. If None, ID of the first plate type (i.e.,
-            96-well) will return
+        name : str, optional
+            Plate type name. If None, the first plate type will be selected
 
         Returns
         -------
@@ -2821,11 +2824,11 @@ class KniminAccess(object):
         Parameters
         ----------
         name : str
-        email : str
-        created_on: datetime
-        notes : str
-        plate_type : str
-            If None, the first plate type (i.e., 96-well) will be used
+        email : str, optional
+        created_on: datetime, optional
+        notes : str, optional
+        plate_type : str, optional
+            If None, the first plate type will be used
 
         Returns
         -------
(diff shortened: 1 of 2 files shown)


knimin/lib/data_access.py
```diff
@@ -2421,14 +2421,14 @@ def _study_exists(self, study_id):
         if not self._con.execute_fetchone(sql, [study_id])[0]:
             raise ValueError('Study ID %s does not exist.' % study_id)
 
-    def _study_is_unique(self, qiita_study_id=None, title=None, study_id=None):
+    def _study_is_unique(self, qiita_study_id=None, title=None, skip_id=None):
         """Confirms that a Qiita study ID and/or a title is not duplicate
 
         Parameters
         ----------
         qiita_study_id : int
```
antgonza: All params are optional, can you add that?

knimin/lib/data_access.py
```diff
@@ -2421,14 +2421,14 @@ def _study_exists(self, study_id):
...
         title : str
-        study_id : int
+        skip_id : int
```
antgonza: Can you also add why/when will we need to pass those params? For example, "useful for testing".

knimin/lib/data_access.py
```diff
@@ -2716,17 +2701,292 @@ def delete_samples(self, ids):
...
+            raise ValueError('Email %s does not exist.' % email)
+
+    def _plate_type_name_to_id(self, name):
+        """Gets plate type ID by name
+
+        Parameters
+        ----------
+        name : str
+            Name of the plate type. If None, ID of the first plate type (i.e.,
+            96-well) will return
```
antgonza: I'm not sure why will you want to return the first plate type? Can you clarify?

knimin/lib/data_access.py
```diff
@@ -2716,17 +2701,292 @@ def delete_samples(self, ids):
...
+            res = self._con.execute_fetchone(sql)
+            return res[0]
+
+    def create_sample_plate(self, name, email=None, created_on=None,
+                            notes=None, plate_type=None):
+        """Creates a new sample plate
+
+        Parameters
+        ----------
+        name : str
```
antgonza: Can you add descriptions and which ones are optional?

knimin/lib/data_access.py
```diff
@@ -2716,17 +2701,292 @@ def delete_samples(self, ids):
...
+            TRN.add(sql, sql_args)
+            return TRN.execute_fetchlast()
+
+    def edit_sample_plate(self, id, name, email=None, created_on=None,
+                          notes=None, plate_type=None):
+        """Edits properties of a sample plate
+
+        Parameters
+        ----------
+        id : int
```
antgonza: Descriptions and optional?

knimin/lib/data_access.py
```diff
@@ -2716,17 +2701,292 @@ def delete_samples(self, ids):
...
+                     WHERE sample_plate_id = %s"""
+            sql_args = [name, email, created_on, notes, plate_type_id, id]
+            TRN.add(sql, sql_args)
+
+    def read_sample_plate(self, id):
+        """Reads properties of an existing a sample plate
+
+        Parameters
+        ----------
+        id : int
```
antgonza: Descriptions?

antgonza: Same for all funcs ...

knimin/lib/data_access.py
qiyunzhu: Here, a study must have either Qiita study ID or title, or both. Therefore, both parameters are optional.

knimin/lib/data_access.py
qiyunzhu: This "skip_id" is for the "edit_study" function but not for the "create_study" function. When the user attempts to rename a study, and types a name that is identical to another study (defined by "skip_id"), an error will be raised. I will add more docstring to explain this.

knimin/lib/data_access.py
qiyunzhu: This is when the user clicks the "Create New Plate" button. I think instead of popping a dialog box asking them which plate type they like, the user may prefer to be directly navigated to a plate map with the default plate type (96-well). This saves one or two user actions. What do you think?

knimin/lib/data_access.py
qiyunzhu: @wasade once commented that some docstrings do not add much information (as the parameter names are simply the column names in the table). Therefore I removed some of them. Should I add "optional" after each optional parameter?

knimin/lib/data_access.py
antgonza: Can you add that explanation to the code?

knimin/lib/data_access.py
antgonza: Thanks.

knimin/lib/data_access.py
antgonza: Well, then what about?
```python
def _plate_type_name_to_id(self, name='96-well'):
   # docs
            sql = """SELECT plate_type_id
                     FROM pm.plate_type
                     WHERE name = %s"""
            res = self._con.execute_fetchone(sql, [name])
            if not res:
                raise ValueError('Plate type %s does not exist.' % repr(name))
            return res[0]
```

knimin/lib/data_access.py
antgonza: parameter names == column names: I think that make sense once you know that but not before. Thus, I think it's worth adding.

optional: yes.

knimin/lib/data_access.py
qiyunzhu: I think your solution works and is more precise, but it assumes that there is always a plate named "96-well", which, may not always be the case in the future.

knimin/lib/data_access.py
qiyunzhu: Got it! Will edit.

wasade:
@qiyunzhu, there appear to be valid test failures

@antgonza's review covers my concerns/questions

knimin/lib/data_access.py
qiyunzhu: Well, I kind of hesitate. The word "96-well" does not even exist in the current database (patch #40). It only occurs in the unit test module. Although I guess that the wet lab will likely add it as the first plate type. If we choose not to mention "96-well" in the function, we are only at the cost of a few more Python lines.
Maybe there is a third solution, that handles this situation in one SQL command. Something like (I am not sure):
`SELECT plate_type_id FROM pm.plate_type WHERE (1 OR (%s IS NOT Null AND name = %s))`

antgonza: What about asking them?

knimin/lib/data_access.py
qiyunzhu: Can you give me an example of writing a docstring to indicate "parameter names == column names"? Thanks!

antgonza: I was listing the different topics in your previous post:
topic 1: parameter names == column names
topic 2: optional

Anyway, the only way I know is to actually write the explanation.

qiyunzhu: I felt that it is not easy to write informative docstring in this scenario. Here is what I came up with:
```
    def edit_study(self, study_id, qiita_study_id=None, title=None, alias=None, notes=None):
        """Edits properties of an existing study

        Parameters
        ----------
        study_id : int
            ID of the study to edit
        qiita_study_id : int, optional
            Assign a Qiita ID to the study
        title : str, optional
            Assign a title to the study
            Either qiita_study_id or title must be specified as the identifier of a study
        alias : str, optional
            Assign an alias to the study
        notes : str, optional
            Assign notes to the study
        """
```
Looks bit redundant and awkward. Do you have some suggestions?

antgonza: I think it looks fine. IMOO It looks redundant to you cause you wrote it but for somebody that hasn't seen that code, the main reason you are writing those docstrings, it's informative.

qiyunzhu:
Hi @wasade thanks for reviewing! Wish you have had good time in Utah :)
I solved the test error and handled some of the @antgonza 's comments (see the new PR). There are two questions remained unsolved at this moment: 1) What is best way to write docstring for parameters that are also column names? 2) Is there a more precise way to write the **`_plate_type_name_to_id`** function?

wasade:
Thanks :) was good times.

Reviewing now

knimin/lib/data_access.py
wasade: I agree with @antgonza as long as the text is accurate.

knimin/lib/data_access.py
```diff
@@ -2716,17 +2705,288 @@ def delete_samples(self, ids):
...
+            If the plate type name does not exist
+        """
+        if name:
+            sql = """SELECT plate_type_id
+                     FROM pm.plate_type
+                     WHERE name = %s"""
+            res = self._con.execute_fetchone(sql, [name])
+            if not res:
+                raise ValueError('Plate type %s does not exist.' % repr(name))
+            return res[0]
```
wasade: this `return` only needs to be done once and can be done outside of the `if` block

wasade:
@qiyunzhu I responded to the comment w.r.t. doc strings. I didn't see a comment specifically about `_plate_type_name_to_id`, only saw one real means for semantic compression but its minor. I think I'm :+1: here once @antgonza's comments are resolved




`

qiyunzhu:
Hi all, I have made changes in response to @antgonza and @wasade 's comments. Docstrings are enriched. Plus, I came up with a very simple and perfect solution for the `_plate_type_name_to_id` function.

antgonza:
@wasade gave his 👍 , with mine, this is ready to merge

qiyunzhu:
Thank you!! @antgonza

PR #146, opened 2016-10-04
Plate mapper 04 sample-study association

qiyunzhu:
In this PR I established the connection between samples and studies. Thanks for reviewing!

### DIFF
--- a/knimin/lib/data_access.py
+++ b/knimin/lib/data_access.py
@@ -2426,7 +2426,10 @@ class KniminAccess(object):
                 raise ValueError('Study ID %s does not exist.' % study_id)
 
     def _study_is_unique(self, qiita_study_id=None, title=None, skip_id=None):
-        """Confirms that a Qiita study ID and/or a title is not duplicate
+        """Confirms that a study is unique
+
+        Confirms that the Qiita study ID and/or the title to be assigned to a
+        study is not duplicate.
 
         Parameters
         ----------
@@ -2446,38 +2449,29 @@ class KniminAccess(object):
         ValueError
             If neither value is given, or either value already exists
         """
+        cols = ['qiita_study_id', 'title']
+        vals = [vars()[col] for col in cols]
+        tags = [x.capitalize().replace('_', ' ').replace(' id', ' ID')
+                for x in cols]
+        if all(val is None for val in vals):
+            raise ValueError('Either %s or %s must be specified.'
+                             % (tags[0], tags[1]))
+        dups = [0 for tag in tags]
         with TRN:
-            tags = ['Qiita study ID', 'Title']
-            cols = [tag.lower().replace(' ', '_') for tag in tags]
-            vals = [vars()[col] for col in cols]
-            if all(val is None for val in vals):
-                raise ValueError('Either %s or %s must be specified.'
-                                 % (tags[0], tags[1]))
-            dups = [0 for tag in tags]
-            for i in range(len(vals)):
-                if tags[i]:
-                    sql = """SELECT study_id
-                             FROM pm.study
-                             WHERE {} = %s
-                             AND study_id IS DISTINCT FROM %s
-                             """.format(cols[i])
-                    TRN.add(sql, [vals[i], skip_id])
+            sql = """SELECT study_id
+                     FROM pm.study
+                     WHERE {} = %s
+                     AND study_id IS DISTINCT FROM %s"""
+            errs = []
+            err_msg = "%s %s conflicts with exisiting study %s."
+            for col, val, tag in zip(cols, vals, tags):
+                if val:
+                    TRN.add(sql.format(col), [val, skip_id])
                     res = TRN.execute_fetchflatten()
                     if res:
-                        dups[i] = res[0]
-            if dups[0] and len(set(dups)) <= 1:
-                raise ValueError('%s %s and %s %s conflict with exisiting '
-                                 'study %s.' % (tags[0], repr(vals[0]),
-                                                tags[1], repr(vals[1]),
-                                                dups[0]))
-            else:
-                errs = []
-                for i in range(len(tags)):
-                    if dups[i]:
-                        errs.append('%s %s conflicts with exisiting study %s.'
-                                    % (tags[i], repr(vals[i]), dups[i]))
-                if errs:
-                    raise ValueError(' '.join(errs))
+                        errs.append(err_msg % (tag, repr(val), res[0]))
+            if errs:
+                raise ValueError('\n'.join(errs))
 
     def create_study(self, qiita_study_id=None, title=None, alias=None,
                      notes=None):
@@ -2569,9 +2563,10 @@ class KniminAccess(object):
             return dict(res[0])
 
     def delete_study(self, study_id):
-        """Deletes an existing study, deletes samples that are exclusively
-        associated with it, and dissociates samples that are associated with it
-        and one or more other studies
+        """Deletes an existing study
+
+        Samples exclusively associated with the study will be deleted. Samples
+        associated with it but also with other studies will be disassociated.
 
         Parameters
         ----------
@@ -2655,42 +2650,46 @@ class KniminAccess(object):
                                      ', '.join(sorted(res)))
 
     def _validate_samples(self, samples):
-        """Confirms that the samples to be created or edited have valid
-        barcodes and study IDs
+        """Validate sample properties
+
+        Confirms that the barcodes and study IDs to be assigned to samples are
+        valid.
 
         Parameters
         ----------
         samples : list of dict
             {barcode : str, optional,
-                Assigns a barcode to the sample
+                Barcode to be assigned to the sample
              study_ids : list of int, optional
-                Associates the sample with one or more studies}
+                One or more study IDs to be associateed with the sample}
 
         Raises
         ------
         ValueError
-            If at least on study ID does not exist
+            If one or more study IDs do not exist
             If one or more sample barcodes do not exist
         """
         with TRN:
             study_ids = set().union(*[x['study_ids'] for x in samples
                                       if 'study_ids' in x])
-            for study_id in study_ids:
-                self._study_exists(study_id)
-            barcodes = [x['barcode'] for x in samples if 'barcode' in x]
-            if barcodes:
-                sql = """SELECT barcode
-                         FROM barcodes.barcode
-                         WHERE barcode IN %s"""
-                TRN.add(sql, [tuple(barcodes)])
-                res = set(barcodes) - set([x[0] for x in
-                                          TRN.execute_fetchindex()])
-                if res:
-                    raise ValueError('Barcode(s) %s do not exist.' %
-                                     ', '.join(sorted(res)))
+            barcodes = set([x['barcode'] for x in samples if 'barcode' in x])
+            cols = ['barcode', 'study_id']
+            tabs = ['barcodes.barcode', 'pm.study']
+            vals = [vars()[col + 's'] for col in cols]
+            tags = [x.capitalize().replace('_', ' ').replace(' id', ' ID') +
+                    '(s)' for x in cols]
+            sql = """SELECT {} FROM {} WHERE {} IN %s"""
+            err_msg = '%s %s do not exist.'
+            for col, tab, val, tag in zip(cols, tabs, vals, tags):
+                if val:
+                    TRN.add(sql.format(col, tab, col), [tuple(val)])
+                    res = val - set([x[0] for x in TRN.execute_fetchindex()])
+                    if res:
+                        val_str = ', '.join(str(x) for x in sorted(res))
+                        raise ValueError(err_msg % (tag, val_str))
 
     def create_samples(self, samples):
-        """Creates samples and associate them with studies
+        """Creates samples and associates them with studies
 
         Parameters
         ----------
@@ -2717,16 +2716,16 @@ class KniminAccess(object):
             sample_ids = [x['id'] for x in samples]
             self._samples_exist(sample_ids, exist=False)
             self._validate_samples(samples)
+            sql1 = """INSERT INTO pm.sample (sample_id, is_blank, barcode,
+                                             notes)
+                      VALUES (%s, %s, %s, %s)"""
+            sql2 = """INSERT INTO pm.study_sample (study_id, sample_id)
+                      SELECT study_ids, %s
+                      FROM unnest(%s) study_ids"""
             for sample in samples:
-                sql = """INSERT INTO pm.sample (sample_id, is_blank, barcode,
-                                                notes)
-                         VALUES (%s, %s, %s, %s)"""
-                TRN.add(sql, [sample['id'], sample.get('is_blank', False),
-                              sample.get('barcode'), sample.get('notes')])
-                sql = """INSERT INTO pm.study_sample (study_id, sample_id)
-                         SELECT study_ids, %s
-                         FROM unnest(%s) study_ids"""
-                TRN.add(sql, [sample['id'], sample['study_ids']])
+                TRN.add(sql1, [sample['id'], sample.get('is_blank', False),
+                               sample.get('barcode'), sample.get('notes')])
+                TRN.add(sql2, [sample['id'], sample['study_ids']])
             TRN.execute()
 
     def edit_samples(self, samples):
@@ -2757,22 +2756,22 @@ class KniminAccess(object):
             sample_ids = [x['id'] for x in samples]
             self._samples_exist(sample_ids)
             self._validate_samples(samples)
+            sql1 = """UPDATE pm.sample
+                      SET is_blank = %s, barcode = %s, notes = %s
+                      WHERE sample_id = %s"""
+            sql2 = """DELETE FROM pm.study_sample
+                      WHERE sample_id = %s"""
+            sql3 = """INSERT INTO pm.study_sample (study_id, sample_id)
+                      SELECT study_ids, %s
+                      FROM unnest(%s) study_ids"""
             for sample in samples:
                 if any(x in sample for x in ['is_blank', 'barcode', 'notes']):
-                    sql = """UPDATE pm.sample
-                             SET is_blank = %s, barcode = %s, notes = %s
-                             WHERE sample_id = %s"""
-                    TRN.add(sql, [sample.get('is_blank', False),
-                                  sample.get('barcode'), sample.get('notes'),
-                                  sample['id']])
+                    TRN.add(sql1, [sample.get('is_blank', False),
+                                   sample.get('barcode'), sample.get('notes'),
+                                   sample['id']])
                 if 'study_ids' in sample:
-                    sql = """DELETE FROM pm.study_sample
-                             WHERE sample_id = %s"""
-                    TRN.add(sql, [sample['id']])
-                    sql = """INSERT INTO pm.study_sample (study_id, sample_id)
-                             SELECT study_ids, %s
-                             FROM unnest(%s) study_ids"""
-                    TRN.add(sql, [sample['id'], sample['study_ids']])
+                    TRN.add(sql2, [sample['id']])
+                    TRN.add(sql3, [sample['id'], sample['study_ids']])
             TRN.execute()
 
     def read_samples(self, ids):
@@ -2851,8 +2850,10 @@ class KniminAccess(object):
             TRN.execute()
 
     def get_samples_by_study(self, study_id):
-        """Retrieves IDs of all samples associated with a study, and whether
-        they are also associated with one or more other studies
+        """Gets samples associated with a study
+
+        Retrieves IDs of all samples associated with a study, and whether
+        they are also associated with other studies.
 
         Parameters
         ----------
@@ -3058,7 +3059,7 @@ class KniminAccess(object):
             TRN.execute()
 
     def read_sample_plate(self, id):
-        """Reads properties of an existing a sample plate
+        """Reads properties of a sample plate
 
         Parameters
         ----------
@@ -3084,8 +3085,10 @@ class KniminAccess(object):
             return dict(TRN.execute_fetchindex()[0])
 
     def _sample_plate_layout_exists(self, id):
-        """Checks whether the layout of a sample plate by ID (at least one
-        sample-to-well record) exists
+        """Checks whether the layout of a sample plate exists
+
+        The layout of a sample plate exists when at least one sample-to-well
+        record exists.
 
         Parameters
         ----------
@@ -3173,7 +3176,7 @@ class KniminAccess(object):
             return [dict(x) for x in TRN.execute_fetchindex()]
 
     def delete_sample_plate(self, id):
-        """Deletes an existing sample plate and its layout
+        """Deletes a sample plate and its layout
 
         Parameters
         ----------
--- a/knimin/lib/data_access.py
+++ b/knimin/lib/data_access.py
@@ -2744,6 +2744,12 @@ class KniminAccess(object):
                 if not self._sample_exists(sample_id):
                     raise ValueError('Sample ID %s does not exist.'
                                      % sample_id)
+                res = self.get_studies_associated_with_samples([sample_id])[0]
+                if res:
+                    study_ids = ', '.join([str(x) for x in res])
+                    raise ValueError('Sample ID %s cannot be deleted because '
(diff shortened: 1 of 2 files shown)


qiyunzhu:
@antgonza Thank you!!

qiyunzhu:
@wasade and @josenavas would you like to add some comments? Thanks!

knimin/lib/data_access.py
```diff
@@ -2570,6 +2570,7 @@ def delete_study(self, study_id):
         """
         self._study_exists(study_id)
         with TRN:
+            self._dissociate_all_samples_from_study(study_id)
```
josenavas: Is this safe? I have the feeling that we should not allow to remove the study if it has samples associated. I think of the study as a container of samples, and I have to have all the samples in a container. If I do this, it is possible that I end up with samples not associated to any study, which would be bad.

knimin/lib/data_access.py
```diff
@@ -2735,15 +2736,22 @@ def delete_samples(self, ids):
...
+            If at least one sample is associated with a sample plate
+            In the above two cases, the study / sample plate has to be deleted
+            prior to the deletion of the sample)
         """
         with TRN:
             for sample_id in ids:
                 if not self._sample_exists(sample_id):
                     raise ValueError('Sample ID %s does not exist.'
                                      % sample_id)
+                res = self.get_studies_associated_with_samples([sample_id])[0]
```
josenavas: In which case it is useful to delete a sample? I'm thinking that if we make the sample deletion too easy we will loose information. In any case, I think the logic here is backwards, i.e. it should be possible to delete samples associated to a study but not studies that has samples. We don't want "orphan" samples.

knimin/lib/data_access.py
```diff
@@ -2759,6 +2767,155 @@ def delete_samples(self, ids):
...
+
+        Returns
+        ------
+        bool
+            Whether they are associated
+        """
+        sql = """SELECT EXISTS (SELECT 1 FROM pm.study_sample
+                                WHERE study_id = %s
+                                AND sample_id = %s)"""
+        return self._con.execute_fetchone(sql, [study_id, sample_id])[0]
```
josenavas: Can you use the TRN object?

knimin/lib/data_access.py
```diff
@@ -2759,6 +2767,155 @@ def delete_samples(self, ids):
...
+        sample_ids : list of str
+            IDs of the samples
+
+        Raises
+        ------
+        ValueError
+            If the study ID does not exist
+            If the study and the sample are already associated
+        """
+        self._study_exists(study_id)
```
josenavas: include this check inside the context manager, so the entire operation occurs in a single transaction

knimin/lib/data_access.py
```diff
@@ -2759,6 +2767,155 @@ def delete_samples(self, ids):
...
+            for sample_id in sample_ids:
+                if not self._sample_exists(sample_id):
+                    raise ValueError('Sample ID %s does not exist.'
+                                     % sample_id)
+                if self._study_sample_are_associated(study_id, sample_id):
+                    raise ValueError('Study ID %s and sample ID %s are already'
+                                     ' associated.' % (study_id, sample_id))
+                sql = """INSERT INTO pm.study_sample (study_id, sample_id)
+                         VALUES (%s, %s)"""
+                TRN.add(sql, [study_id, sample_id])
```
josenavas: don't forget the TRN.execute() once the for is completed. Although not needed, it is safe to ensure that once this function is completed, it actually performed the operation that is supposed to do.

knimin/lib/data_access.py
```diff
@@ -2759,6 +2767,155 @@ def delete_samples(self, ids):
...
+
+    def dissociate_samples_from_study(self, study_id, sample_ids):
```
josenavas: What happens if we end up with orphan samples? I know I missed a couple of PRs, but in my opinion we work in a "study" basis rather than in a "sample" basis. I.e. we are not going to process samples if we don't know to which study they belong to.

knimin/lib/data_access.py
```diff
@@ -2759,6 +2767,155 @@ def delete_samples(self, ids):
...
+            for sample_id in sample_ids:
+                if not self._sample_exists(sample_id):
+                    raise ValueError('Sample ID %s does not exist.'
+                                     % sample_id)
+                if not self._study_sample_are_associated(study_id, sample_id):
+                    raise ValueError('Study ID %s and sample ID %s are not '
+                                     'associated.' % (study_id, sample_id))
+                sql = """DELETE FROM pm.study_sample
+                         WHERE study_id = %s AND sample_id = %s"""
+                TRN.add(sql, [study_id, sample_id])
```
josenavas: Same comment as above re `TRN.execute` (last time I point it out, but applies to all)

knimin/lib/data_access.py
```diff
@@ -2759,6 +2767,155 @@ def delete_samples(self, ids):
...
+        ----------
+        study_id : int
+            ID of the study
+
+        Returns
+        ------
+        list of str
+            Sorted list of sample IDs associated with the study
+        """
+        self._study_exists(study_id)
```
josenavas: Please, use the `TRN` object. Applies to all.

knimin/lib/data_access.py
```diff
@@ -2759,6 +2767,155 @@ def delete_samples(self, ids):
...
+        list of list of int
+            List of sorted list of study IDs associated with each sample in
+            the same order as sample IDs
+
+        Raises
+        ------
+        ValueError
+            If at least one sample ID does not exist
+        """
+        associations = []
```
josenavas: All this operation can be performed in a single SQL query, please update (should be orders of magnitude faster)

qiyunzhu:
@josenavas Thanks for reviewing! I will handle your comments soon. - Qiyun

knimin/lib/data_access.py
qiyunzhu: I proposed this question and discussed with @wasade . We think that we should allow a more relaxed association between sample and study. The rationale is that one sample can be shared by multiple studies. Therefore, I made this design, in which samples and studies are independent from each other, until they are associated.

knimin/lib/data_access.py
qiyunzhu: Got it, thanks!

knimin/lib/data_access.py
qiyunzhu: See my comment below.

knimin/lib/data_access.py
qiyunzhu: See my comment below.

knimin/lib/data_access.py
josenavas: @qiyunzhu my comment is about the location of this line (move one line below):
```python
with TRN:
    self._study_exists(study_id)
```
that would make all the SQL interactions a single transaction

knimin/lib/data_access.py
josenavas: If I remember the discussion correctly, was to allow a sample to be associated with multiple studies. The way I interpret that, is that it changes the upper limit of the number of studies that a sample can be associated (from 1 to many). However, I think it doesn't change the lower limit, which is that each sample should be at least associated with a study. cc @wasade @antgonza

knimin/lib/data_access.py
josenavas: I think this entire function can be replaced with (probably a couple of tweeks with the `association` variable to match types:
```python
sql = """SELECT sample_id, array_agg(study_id ORDER BY study_id)
         FROM pm.study_sample
         WHERE sample_id IN (%s)"""
TRN.add(sql, [tuple(sample_ids])
associations = TRN.execute_fetchindex()
```

Any special reason for returning a list of list of int rather than a dict of `{sample_id: [study_ids]}`. I have the feeling that relying on sorting is fragile.

knimin/lib/data_access.py
antgonza: I think the 1 -> many makes sense and not allowing having 0 studies related to sample makes sense. However, I'm not sure how complicated solving that issue is so if is hard, I'm OK with opening an issue.

qiyunzhu: @antgonza The implementation is also among my concerns. Also, is it absolutely not possible that some samples are out there without an associated study? Could it happen that the wet lab people found that one sample is mistakenly assigned to a study. Then they have to unlink the study, but they cannot figure out immediately which study this sample actually belongs to. So they would rather leave this empty for the moment. Therefore, allowing it to be 0 gives people more flexibility. @josenavas If "person" and "date" of a plate can be empty (as you suggested), I think "study" of a sample can also be empty.

antgonza: "one sample is mistakenly assigned to a study", it could happen (anything could happen) but if it happens then we should not only delete from the DB but also not process as a "floating" sample is useless, at least IMOO.

qiyunzhu:
We can delete a bad sample in silico, but in the real world it is already
physically present in a sample / DNA plate, along with many other good
samples. The wet lab crew cannot delete the sample by draining the well
with a pipettor, or pretend that the well is empty. I guess that the only
real-world method to delete a sample is to dump the entire sample plate.

knimin/lib/data_access.py
qiyunzhu: We can delete a bad sample in silico, but in the real world it is already physically present in a sample / DNA plate, along with many other good samples. The wet lab crew cannot delete the sample by draining the well with a pipettor, or pretend that the well is empty. I guess that the only real-world method to delete a sample is to dump the entire sample plate.

antgonza:
Replaying here cause you replied to this thread. I guess you can always change the name to be BLANK. Anyway, I think the best way to go about this is to ask the wetlab how often this happens to figure out if is a real concern. I _really_ hope that this never happens.

knimin/lib/data_access.py
qiyunzhu: Also, to biologists samples are more precious than anything. When they come across an eppendorf tube with missing label during freezer cleanup, they will most likely still keep it as if they can remember what it is in the future.

antgonza: Thanks. I agree to go with whatever the final user decides is best 😄 .

qiyunzhu: Marking obsolete wells as BLANK may be a good solution. @josenavas once proposed to have "is_blank" as a property of a sample. This may be the solution, as samples with is_blank=True doesn't have to be associated with a study. Anyway let's see how the wet lab thinks.

qiyunzhu: I totally agree!

knimin/lib/data_access.py
qiyunzhu: By the way, it there a `fetchdict` method in TRN?

knimin/lib/data_access.py
qiyunzhu: The function does not sort. It simply returns a list of [study_ids] in the same order as the input sample_ids. The "ORDER BY" clause you see is for each child [study_ids]. It is not absolute necessary.

antgonza: Suggest running in python ...

qiyunzhu: Hi @wasade Thanks for the input. I tried but still no luck. I did:
`SELECT sample_plate_id, array_agg(sample_id) FROM pm.sample_plate_layout WHERE sample_plate_id IN (1, 2, 3);`
The screen output is:
`ERROR:  column "sample_plate_layout.sample_plate_id" must appear in the GROUP BY clause or be used in an aggregate function`
Sorry for asking such a concrete question. I already Googled but didn't get a solution on the first hand. Therefore I come to bother you here.

antgonza: SELECT sample_plate_id, array_agg(sample_id) FROM pm.sample_plate_layout WHERE sample_plate_id IN (1, 2, 3) GROUP BY sample_plate_id;

knimin/lib/data_access.py
qiyunzhu: Hi @josenavas May I double-confirm with you, that do you want all SQL operations to be done using TRN, even if the operation does not write the database?

antgonza: That's normally what we do ...

qiyunzhu: @antgonza Got it! Will do.

knimin/lib/data_access.py
qiyunzhu: @antgonza Sure! Thanks! There is no more reply from others so far.

wasade: @qiyunzhu multiple is somewhat weird as, for an exists, you need to decide if it returns true if *any* samples exist or if and only if *all* samples exist

qiyunzhu: @wasade I know, but I think @josenavas 's original idea is to combine multiple SQL commands into one. If we still have to iterate sample IDs then the performance improvement will be marginal. I think the `samples_exist` function can return a dict of `{sample_id: true/false}`. @josenavas what do you think?

qiyunzhu:
Hi all,

In order to implement the mechanism that a sample must be associated with one or more studies, I spent long time to work it out, and the changes are substantial. Please excuse me if you find there are too many lines edited, but it can't be any shorter in order to produce a functional, logically intact unit.

Here is the explanation of the new mechanism:

1. Since a sample must be associated with at least one study, I made the program to work in a way as if "study IDs" is a property of a sample (though actually this information is in a separate table). The sample-related functions: `create/edit/read/delete_samples`, now take care of the association and dissociation of studies. Meanwhile, the "association" functions I created in my last version of this PR, are all removed.

2. I optimized the code to the best of my knowledge, so that least number of SQL commands has to be executed to complete those bulk operations related to samples. (Maybe there are better ways, though)

(to be continued as I have to leave)

knimin/lib/data_access.py
```diff
@@ -2443,35 +2446,38 @@ def _study_is_unique(self, qiita_study_id=None, title=None, skip_id=None):
...
-                                      repr(vals[1]), dups[0]))
-        else:
-            errs = []
-            for i in range(len(tags)):
-                if dups[i]:
-                    errs.append('%s %s conflicts with exisiting study %s.'
-                                % (tags[i], repr(vals[i]), dups[i]))
-            if errs:
-                raise ValueError(' '.join(errs))
+        with TRN:
```
josenavas: This entire function can be substituted by:
```python
def _study_is_unique(self, qiita_study_id=None, title=None, skip_id=None):
    """DOCS..."""
    if qiita_study_id is None and title is None:
        raise ValueError('Either qiita_study_id or title must '
                         'be specified')

    with TRN:
        cols = ['qiita_study_id', 'title']
        vals = [qiita_study_id, title]
        sql = """SELECT study_id
                 FROM pm.study
                 WHERE {} = %s AND study_id IS DISTINCT FROM %s"""

        errors = []
        err_msg = "Study with %s '%s' already exists: %s"
        for col, val in zip(cols, vals):
            TRN.add(sql.format(col), [val, skip_id])
            res = TRN.execute_fetchflatten()
            if res:
                errors.append(err_msg % (col, val, res[0]))

        if errors:
            raise ValueError('\n'.join(errors))
```

Which in my opinion is more readable and easy to maintain. Also, need to update the documentation to reflect that there is another case in which ValueError can be raised

knimin/lib/data_access.py
```diff
@@ -2588,110 +2615,168 @@ def _sample_exists(self, sample_id):
...
-            If the barcode does not exist
+            If at least on study ID does not exist
+            If one or more sample barcodes do not exist
         """
-        sql = """SELECT EXISTS (SELECT 1 FROM barcodes.barcode
-                                WHERE barcode = %s)"""
-        if not self._con.execute_fetchone(sql, [barcode])[0]:
-            raise ValueError('Barcode %s does not exist.' % barcode)
+        with TRN:
+            study_ids = set().union(*[x['study_ids'] for x in samples
```
josenavas:
```python
from itertools import chain

study_ids = set(chain.from_iterable([x['study_ids'] for x in samples if 'study_ids' in x]))
```

knimin/lib/data_access.py
```diff
@@ -2588,110 +2615,168 @@ def _sample_exists(self, sample_id):
...
+                                      if 'study_ids' in x])
+            for study_id in study_ids:
```
josenavas: I think it would be better to test all the studies at once, as you're doing for the barcodes, to minimize the number of SQL calls and parallelize checks. This way, the user may have a list of 10 studies (that none of them exist), check the first one, fail, then check the next one, fail, etc... up to 10 times. If done like the barcodes test, all the checks happen at once and the user only see 1 error message with all the problems that he has. Does it make sense?

knimin/lib/data_access.py
```diff
@@ -2588,110 +2615,168 @@ def _sample_exists(self, sample_id):
...
+
+    def _validate_samples(self, samples):
+        """Confirms that the samples to be created or edited have valid
+        barcodes and study IDs
+
+        Parameters
+        ----------
+        samples : list of dict
+            {barcode : str, optional,
+                Assigns a barcode to the sample
```
josenavas: Update documentation, this function is not assigning/associating anything, is just validating.

knimin/lib/data_access.py
```diff
@@ -2588,110 +2615,168 @@ def _sample_exists(self, sample_id):
...
-                sample_id = sample['id']
-                if self._sample_exists(sample_id):
-                    raise ValueError('Sample ID %s already exists.'
-                                     % sample_id)
-                barcode = sample.get('barcode')
-                if barcode:
-                    self._barcode_exists(barcode)
-                is_blank = sample.get('is_blank', False)
-                notes = sample.get('notes')
                 sql = """INSERT INTO pm.sample (sample_id, is_blank, barcode,
```
josenavas: Minor, but you can specify this SQL string outside the for loop, so you don't have to create it every time.

knimin/lib/data_access.py
```diff
@@ -2588,110 +2615,168 @@ def _sample_exists(self, sample_id):
...
                                                 notes)
-                         VALUES (%s, %s, %s, %s)
-                         RETURNING sample_id"""
-                TRN.add(sql, [sample_id, is_blank, barcode, notes])
+                         VALUES (%s, %s, %s, %s)"""
+                TRN.add(sql, [sample['id'], sample.get('is_blank', False),
+                              sample.get('barcode'), sample.get('notes')])
+                sql = """INSERT INTO pm.study_sample (study_id, sample_id)
```
josenavas: Same here (just provide different names to the `sql` variable)

knimin/lib/data_access.py
```diff
@@ -2588,110 +2615,168 @@ def _sample_exists(self, sample_id):
...
-                    self._barcode_exists(barcode)
-                is_blank = sample.get('is_blank', False)
-                notes = sample.get('notes')
-                sql = """UPDATE pm.sample
-                         SET is_blank = %s, barcode = %s, notes = %s
-                         WHERE sample_id = %s
-                         RETURNING sample_id"""
-                TRN.add(sql, [is_blank, barcode, notes, sample_id])
+                if any(x in sample for x in ['is_blank', 'barcode', 'notes']):
+                    sql = """UPDATE pm.sample
```
josenavas: Same comment as above regarding SQL variables declaration (outside the loop)

knimin/lib/data_access.py
```diff
@@ -2906,18 +3043,19 @@ def edit_sample_plate(self, id, name, email=None, created_on=None,
             Defines the plate type of the sample plate
             If None, the first plate type will be used
         """
-        self._sample_plate_exists(id)
-        self._sample_plate_is_unique(name, id)
-        if email:
-            self._email_exists(email)
-        plate_type_id = self._plate_type_name_to_id(plate_type)
         with TRN:
+            self._sample_plate_exists(id)
```
josenavas: I'm just gonna mention it here although there is no necessity to change at this moment. Usually is a bad idea to use variable names that match functions defined in python. `id` is a function in python so it will probably be a better idea to name the variable `plate_id`. In this function, you no longer have access to the python function `id`, and although you're not using it, it is usually a better practice not to overwrite it. This is just FYI, and no need to change now, although it will be nice to eventually change it.

knimin/lib/data_access.py
josenavas: As @wasade says, it depends on the function definition - the dict {true/false} is the more general solution, although if you are sure which is the behavior that you want you can do _all_samples_exist and _any_samples_exist functions to have both behaviors. Readability is one of the most important things.

knimin/lib/data_access.py
```diff
@@ -2443,35 +2449,28 @@ def _study_is_unique(self, qiita_study_id=None, title=None, skip_id=None):
         ValueError
             If neither value is given, or either value already exists
         """
-        tags = ['Qiita study ID', 'Title']
-        cols = [tag.lower().replace(' ', '_') for tag in tags]
+        cols = ['qiita_study_id', 'title']
         vals = [vars()[col] for col in cols]
```
josenavas: Any specific reason why you keep the call to `vars()`? That is one of the changes that I proposed in my version and the reason behind is that the current way of doing it becomes not that readable.

This version:
```
vals = [qiita_study_id, title]
```
is more readable and easy to maintain... There is no reason for making this call so "meta"

knimin/lib/data_access.py
```diff
@@ -2443,35 +2449,28 @@ def _study_is_unique(self, qiita_study_id=None, title=None, skip_id=None):
...
-                if dups[i]:
-                    errs.append('%s %s conflicts with exisiting study %s.'
-                                % (tags[i], repr(vals[i]), dups[i]))
+            err_msg = "%s %s conflicts with exisiting study %s."
+            for col, val, tag in zip(cols, vals, tags):
+                if val:
+                    TRN.add(sql.format(col), [val, skip_id])
+                    res = TRN.execute_fetchflatten()
+                    if res:
+                        errs.append(err_msg % (tag, repr(val), res[0]))
```
josenavas: Why calling `repr(val)`? `val` is going to be either a string or an integer, both of them can be used directly.

knimin/lib/data_access.py
```diff
@@ -2588,110 +2609,172 @@ def _sample_exists(self, sample_id):
...
-        sql = """SELECT EXISTS (SELECT 1 FROM barcodes.barcode
-                                WHERE barcode = %s)"""
-        if not self._con.execute_fetchone(sql, [barcode])[0]:
-            raise ValueError('Barcode %s does not exist.' % barcode)
+        with TRN:
+            sql = """SELECT sample_id
+                     FROM pm.sample
+                     WHERE sample_id IN %s"""
+            TRN.add(sql, [tuple(sample_ids)])
+            res = [x[0] for x in TRN.execute_fetchindex()]
```
josenavas: I'm unsure, but I think `res = TRN.execute_fetchflatten()` will given you the same result in this call.

knimin/lib/data_access.py
```diff
@@ -2588,110 +2609,172 @@ def _sample_exists(self, sample_id):
...
+            tabs = ['barcodes.barcode', 'pm.study']
+            vals = [barcodes, study_ids]
+            tags = [x.capitalize().replace('_', ' ').replace(' id', ' ID') +
+                    '(s)' for x in cols]
+            sql = """SELECT {} FROM {} WHERE {} IN %s"""
+            err_msg = '%s %s do not exist.'
+            for col, tab, val, tag in zip(cols, tabs, vals, tags):
+                if val:
+                    TRN.add(sql.format(col, tab, col), [tuple(val)])
+                    res = val - set([x[0] for x in TRN.execute_fetchindex()])
```
josenavas: `res = val - set(TRN.execute_fetchflatten())`

wasade:
I don't have further comments on this PR

knimin/lib/data_access.py
qiyunzhu: Thanks for telling me this trick. Tested and working. Edited.

knimin/lib/data_access.py
qiyunzhu: Same as above.

knimin/lib/data_access.py
qiyunzhu: I changed this in your way.

knimin/lib/data_access.py
qiyunzhu: That's because `repr` will add quotes to a string but not to a number. When applying to ID and title, the outcome will read like: `Warning: Sample ID 123 'my favorite sample' is not valid.`

Issue #145, opened 2016-10-03
"Action Required - Assign Sample" e-mail has outdated instructions 


BrentKLer:
Hi, 

We are sending out these e-mails to participants who have sent in samples but not assigned them but the language in the e-mail is from our old instructions and many are getting confused. Hoping someone could update the e-mail to reflect the new instructions. Thanks!

Currently the e-mail says "We have recently received your sample barcode: x, but we cannot process
your sample until the following steps have been completed online. Please ensure
that you have completed both steps outlined below:
1. Submit your consent form and survey

Consent and survey depend on sample type. For human samples, the consent form
is mandatory. Even if you elect not to answer the questions on the survey,
please click through and submit the survey in order to ensure we receive your
completed consent form.
For pet samples, we ask that you fill out a short survey. No consent form is
necessary.
For environmental samples, the consent form and survey are not necessary.

To begin the consent/survey process:
- Click on the "Add Source Survey" tab on the main page.
- Select the appropriate category (human, animal, or environmental) for your
  sample.
1. Associate your sample(s) with your survey(s)

This step is important as it connects your consent form to your sample. We
cannot legally work with your sample until this step has been completed. For
human and pet samples, the survey must be completed before doing this step.

To associate your sample with your survey:
- Log into your account and click the "Associate/Log Sample" button at the
  bottom of the left-hand navigation menu. This will bring you to a screen
  with the heading "Choose your sample source".
- Click on the name of the participant that the sample belongs to.
- Fill out the required fields and submit.

The American Gut participant website is located at
https://microbio.me/americangut
The British Gut participant website is located at
https://microbio.me/britishgut
If you have any questions, please contact us at info@americangut.org.

Thank you,
American Gut Team"

EmbrietteH:
Just worked with Brent on this. Proposed changes:

We have recently received your sample barcode: x, but we cannot process your sample until the following steps have been completed online. Please ensure that you have completed both steps outlined below:

1. **Submit your consent form and survey-_if you have already done these please proceed to step 2 below._**
For human samples, the consent form is mandatory. Even if you elect not to answer the questions on the survey, please click through and submit the survey in order to ensure we receive your completed consent form.

**2. Assign your sample(s) to your survey(s)**
This step is critical as it connects your consent form to your sample. We
cannot legally work with your sample until this step has been completed. 

To assign your sample to your survey:

[...]

PR #148, opened 2016-10-05
Update email text

sjanssen2:
See issue  #145. Who is going to roll out these changes into the productive system, such that the changed emails can be sent out by Brent?


wasade:
test failure :( looks like there is a pretty stringent check on the content of the email

PR #149, opened 2016-10-06
Let's try this. py2 utf-8 handling is so broken.

wasade:
@EmbrietteH @sjanssen2 @josenavas this I think solves the utf-8 issue from this morning.


sjanssen2:
looks good to me. Am I right that we now have utf-8 'state's? What about the other fields, like 'name', 'address', 'city', ... ?

wasade:
We do. I modified so geocode information is always decoded. Participant names should already be. What we really need to do is get away from py2 as its handling of ascii and utf-8 is horrible.

EmbrietteH:
Thanks both of you! Good to merge?

sjanssen2:
:+1: good to merge.

PR #157, opened 2016-10-24
changed test for PDF generation

sjanssen2:
I am trying to test if a PDF file has been generated. I am checking for some strings in the text representation of the PDF. It turns out that the content changes with used font. Thus, I should no longer check for the string "Length 10559" but for something stable.
Now, tests should pass again.


wasade:
Thanks!

Issue #158, opened 2016-10-28
A scanned barcode should note host type


wasade:
eg human, non-human, environmental

sjanssen2:
Could you be more specific to enable me to work on that. 
Is this information already stored in the DB?
On which pages (URLs, Handlers) should that be displayed?

wasade:
It should be based off the information from logging the sample (e.g.,
human, animal or environmental). The specific place where it should be
displayed is upon scanning a barcode when it is received

sjanssen2:
I never saw labadmin in action, thus I am not sure which handler will be used to enable scanning barcodes :-/

wasade:
Should be here:
https://github.com/biocore/labadmin/blob/master/knimin/handlers/barcode_util.py

wasade:
It might make sense to chat at somepoint with the wetlab to see what they
do when samples come in.

sjanssen2:
a note to myself: according to Greg "Lindsay DeRight Goldasich" is the person who scans all the barcodes.

sjanssen2:
@wasade who came up with this request?
Btw: could you review PR #159 to let me move on here.
Thank!

wasade:
I made the request as I thought the barcode page was already showing this distinction, and was surprised when Lindsay said it wasn't

sjanssen2:
aha. Maybe I am too stupid to find it, but were is the information of the "host type" stored in our DB? Do we have to infer that from the survey taken?

wasade:
It's not exactly host type. The distinction is human, animal or
environmental. IIRC, this is inferred and not explicit in the database. I'd
need to search around, although I _think_ there is some detail in
`ag_kit_barcodes`

sjanssen2:
hope to address this by PR #165

PR #159, opened 2016-10-31
added new unit tests for the get method of handler "BarcodeUtilHandler"

sjanssen2:
In order to address #158 I'd like to have unit tests in place. I think I found some odd things in the code and will correct them after the test PR has been accepted. Until now, I haven't touched the handler sources other than adding TODOs.

### DIFF
--- a/knimin/handlers/barcode_util.py
+++ b/knimin/handlers/barcode_util.py
@@ -201,6 +201,9 @@ class BarcodeUtilHandler(BaseHandler, BarcodeUtilHelper):
         project_names = db.getProjectNames()
 
         # barcode exists get general info
+        # TODO (Stefan Janssen): check spelling of "received", i.e. tests in
+        # the template check for 'Recieved'. I think the logic is broken due
+        # to that.
         if barcode_details['status'] is None:
             barcode_details['status'] = 'Received'
         if barcode_details['biomass_remaining'] is None:
@@ -213,6 +216,8 @@ class BarcodeUtilHandler(BaseHandler, BarcodeUtilHelper):
         if (barcode_details['obsolete'] == "Y"):
                 # the barcode is obsolete
                 div_id = "obsolete"
+                # TODO: Stefan: why is that set here, as far as I see, this
+                # message will in all cases be overwritten!
                 message = "Barcode is Obsolete"
         # get project info for div
         ag_details = []
--- a/knimin/tests/test_ag_pulldown.py
+++ b/knimin/tests/test_ag_pulldown.py
@@ -9,7 +9,7 @@ from knimin import db
 
 
 class testUpdateEBIStatusHandler(TestHandlerBase):
-    os.environ["ASYNC_TEST_TIMEOUT"] = "30"
+    os.environ["ASYNC_TEST_TIMEOUT"] = "60"
 
     def test_get_not_authed(self):
         response = self.get('/update_ebi/')
--- a/knimin/tests/test_ag_pulldown.py
+++ b/knimin/tests/test_ag_pulldown.py
@@ -9,7 +9,7 @@ from knimin import db
 
 
 class testUpdateEBIStatusHandler(TestHandlerBase):
-    os.environ["ASYNC_TEST_TIMEOUT"] = "60"
+    os.environ["ASYNC_TEST_TIMEOUT"] = "30"
 
     def test_get_not_authed(self):
         response = self.get('/update_ebi/')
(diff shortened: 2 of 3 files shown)


knimin/handlers/barcode_util.py
```diff
@@ -201,6 +201,9 @@ def get(self):
         project_names = db.getProjectNames()
 
         # barcode exists get general info
+        # TODO (Stefan Janssen): check spelling of "received", i.e. tests in
```
wasade: what??

knimin/handlers/barcode_util.py
```diff
@@ -213,6 +216,8 @@ def get(self):
         if (barcode_details['obsolete'] == "Y"):
                 # the barcode is obsolete
                 div_id = "obsolete"
+                # TODO: Stefan: why is that set here, as far as I see, this
```
wasade: I concur

knimin/tests/test_barcode_util.py
```diff
@@ -281,5 +284,239 @@ def test_build_email(self):
...
+        response = self.get('/barcode_util/',
+                            {'barcode': barcode_biomass_empty})
+        self.assertEqual(response.code, 200)
+        self.assertIn('<option value="N" >No</option>', response.body)
+        self.assertIn('<option value="Y" >Yes</option>', response.body)
+        self.assertIn('<option value="Unknown" selected>Unknown</option>',
+                      response.body)
+
+        # check if sequencing status is set to ''
+        # (TODO: currently it is set to WAITING, which seems to be wrong!)
```
wasade: Why is WAITING wrong?

wasade:
A flake8 error and it looks like a timeout again :/

knimin/handlers/barcode_util.py
sjanssen2: In the DB, it is written `"Received"`, but in the template it is used like
`{% for bstatus in ['Recieved', ''] %}
    {% if bstatus == barcode_info['status'] %}`
So I don't see how that condition can ever be satisfied.

knimin/tests/test_barcode_util.py
sjanssen2: We have the following values in our DB for the column sequencing_status:
"''"
""
"FAILED_SEQUENCING"
"FAILED_SEQUENCING_1"
"FAILED_SEQUENCING_2"
"SUCCESS"
"WAITING"
Thus, I figure there should be a difference between a barcode marked as "waiting" and a barcode without any information, especially since there is the missing value option in the combo box `<option value="" ></option>`

wasade: Agree, thanks

knimin/handlers/barcode_util.py
wasade: Yea... that's not good.

knimin/tests/test_barcode_util.py
```diff
@@ -518,6 +519,142 @@ def test_get(self):
...
+        del data['scan_date']
+        response = self.post('/barcode_util/', data)
+        self.assertEqual(response.code, 200)
+        dbInfo = db.get_barcode_details(data['barcode'])
+        self.assertEqual(None, dbInfo['sample_postmark_date'])
+        self.assertEqual(None, dbInfo['scan_date'])
+        data['postmark_date'] = '2015-06-25'
+        data['scan_date'] = '2015-07-01'
+
+        # TODO: Stefan Janssen: looks like we can successfully update a non
```
wasade: :(

knimin/tests/test_barcode_util.py
```diff
@@ -518,6 +519,142 @@ def test_get(self):
...
+
+        # check that update failes for wrong data types
+        data['postmark_date'] = 'invalid date'
+        response = self.post('/barcode_util/', data)
+        self.assertEqual(response.code, 200)
+        self.assertIn("Barcode %s general details failed" % data['barcode'],
+                      response.body)
+
+        # test changing the barcode's project to a non existing one
+        # TODO: Stefan Janssen: I think this should not result in a positive
```
wasade: agree

wasade:
Just one question about the `599`

knimin/tests/test_ag_pulldown.py
```diff
@@ -21,13 +21,14 @@ def test_get_not_authed(self):
...
         self.mock_login_admin()
-        os.environ["ASYNC_TEST_TIMEOUT"] = "180"
+        os.environ["ASYNC_TEST_TIMEOUT"] = "30"
 
         # test successful query
         response = self.get('/update_ebi/')
-        self.assertEqual(response.code, 200)
-        self.assertIn('Successfully updated barcodes in database',
-                      response.body)
+        self.assertIn(response.code, [200, 599])  # either success, or time out
```
sjanssen2: I have the feeling that Travis imposes some limitations on the time-out length, because it always returns fine, i.e. with 200, when running this test on my local machine. I don't want to spend too much effort here to figure out how to make this happen within Travis. 

But any recommendations are welcome!

wasade: Okay, sounds good. On closer [inspection](https://github.com/biocore/labadmin/blob/5861c375950c175499bfe17ba29b02c3d092927a/knimin/lib/data_access.py#L2399-L2413), it could also be the case that EBI is throttling travis as well. I don't think there is an easy solution here.

PR #163, opened 2016-11-30
Pulldown: select surveys to be exported + merged table

sjanssen2:
1. Update to the metadata pulldown handler, such that the user now sees which different internal surveys are available plus he/she can select which of those should be exported to the download.

2. There is now an option to join all columns from all surveys into one huge file.

3. Filenames in resulting zip archive are more speaking now.

4. There seems to be a change in the flake8 on Travis. Thus, I had to add some blank lines to certain files.


knimin/lib/tests/test_data_access.py
```diff
@@ -297,6 +297,20 @@ def test_get_ag_barcode_details(self):
...
+                               (-4, 'Surfers', True),
+                               (-5, 'Personal_Microbiome', True)],
+                              db.list_ag_surveys())
+        self.assertItemsEqual([(-1, 'Personal Information', False),
+                               (-2, 'Pet Information', True),
+                               (-3, 'Fermented Foods', False),
+                               (-4, 'Surfers', True),
+                               (-5, 'Personal_Microbiome', False)],
+                              db.list_ag_surveys([-2, -4]))
+
```
mortonjt: Just to clarify - this is testing the real database if these surveys in the db?

mortonjt: I do think that this is a really critical check.  Now it looks like every time that a survey gets added, this test would need to be updated.  What do you think about this?  Do you think that it is worth while to raise an issue to make these tests - so that we don't need to update these tests every time that the surveys are updated.

sjanssen2: I don't think that adding new surveys is a frequently occurring task. Thus, updating the test each time should be fine.

knimin/templates/ag_pulldown.html
```diff
@@ -71,13 +80,33 @@ <h3 style="color:red">Pulldown Processing, please wait for file download. It may
 <form enctype="multipart/form-data" action="/ag_pulldown/" name="agForm" id="agForm" method="post">
 <p>Upload qiime mapping file, or other file with first line header and one barcode per line</p>
 <p>Barcodes File <input type="file" name="barcodes" id="barcodes"/></p>
-<p>External Survey
+<p>American Gut Surveys:
+  <div>Select which American Gut surveys shall be included in the pulldown archive. <br/>Add 'all in one file' if you also want to have one addition file<br/>that contains all available columns from all American Gut surveys.</div>
+  <select name="agsurveys" id="agsurveys" multiple />
+  {% for s in agsurveys %}
+  {% if s[2] == True %}
```
wasade: `if s[2]`

knimin/templates/ag_pulldown.html
```diff
@@ -71,13 +80,33 @@ <h3 style="color:red">Pulldown Processing, please wait for file download. It may
...
+</p>
+<p>External Survey:
+  <div>Select which external surveys should be included in the pulldown archive.</div>
   <select name="external" id="external" multiple />
   {% for s in surveys %}
   <option value='{{s}}'>{{s}}</option>
   {% end %}
   </select>
 </p>
+{% if merged == 'True' %}
```
wasade: this is a string comparison?

sjanssen2: yes it is, since 'merged' holds the content of 'value' of the checkbox which I defined as the string 'True'.

knimin/templates/ag_pulldown.html
sjanssen2: fixed

wasade:
@sjanssen2 looks like there are merge conflicts

sjanssen2:
We have issues with Geocoder searching for locations where addresses are now random utf-8 characters. This results in unstable results for metadata pulldown. Thus, I only compare the first 1000 characters of each file in the pulldown archive - basically to continue working on #184 
Let us fix Geocoder later.

sjanssen2:
@wasade @josenavas @antgonza can I please get a rapid review to be able to continue working on #184

knimin/handlers/ag_pulldown.py
```diff
@@ -67,8 +84,42 @@ def post(self):
...
-            meta_zip.append('survey_%s_md.txt' % survey, meta)
+            # only create files for those surveys that have been selected by
+            # the user. Note that ids from the DB are negative, in metadata
+            # they are positive!
+            # Currently, I (Stefan Janssen) don't have test data for external
+            # surveys, thus I don't know their 'survey' value. I expect it to
+            # be the name of the external survey. In order to not block their
+            # pulldown I check that a skipped survey ID must be in the set of
+            # all available surveys.
+            if ((-1 * survey) in selected_ag_surveys) or \
```
josenavas: `abs_survey = abs(survey)` and use `abs_survey` instead of `-1 * survey` in order to avoid code duplication

knimin/handlers/ag_pulldown.py
```diff
@@ -67,8 +84,42 @@ def post(self):
...
+                                available_agsurveys[-1 * survey], meta)
+                # transform each survey into a pandas dataframe for later merge
+                # read all columns as string to avoid unintened conversions,
+                # like cutting leading zeros of barcodes
+                pd_meta = pd.read_csv(StringIO(meta), sep="\t", dtype=str)
+                # reset the index to barcodes = here sample_name
+                pd_meta.set_index('sample_name', inplace=True)
+                results_as_pd.append(pd_meta)
+
+        pd_all = pd.DataFrame()
```
josenavas: Should these 3 lines of code be moved inside the `if` below? Not sure how expensive is the merge, but if it is only needed when merged = True then just do it there.

knimin/lib/data_access.py
```diff
@@ -1685,6 +1685,34 @@ def add_external_survey(self, survey, description, url):
                  RETURNING external_survey_id"""
         return self._con.execute_fetchone(sql, [survey, description, url])[0]
 
+    def list_ag_surveys(self, selected=None):
+        """Returns the list of american gut survey names
+
+        Parameters
+        ----------
+        selected : list of int
+            To keep track of which surveys have been selected by the user via
```
josenavas: `the resulting list of tupels third element holds this information as a boolean value` -> I don't understand what this means, can you rephrase?

knimin/lib/data_access.py
```diff
@@ -1685,6 +1685,34 @@ def add_external_survey(self, survey, description, url):
...
+        list of (int, str, bool)
+            first element is the group_order number
+            second element is the group name
+            third element is the information if user selected this survey for
+                  pulldown. All surveys are selected if selected is None.
+        """
+        sql = """SELECT group_order, american
+                FROM ag.survey_group
+                WHERE group_order < 0"""
+        return [(id, name, (selected is None) or (id in selected))
```
josenavas: `id` -> `id_` 
`id` is a reserved word

knimin/lib/data_access.py
```diff
@@ -1685,6 +1685,34 @@ def add_external_survey(self, survey, description, url):
...
+            the rendered website, the resulting list of tupels third element
+            holds this information as a boolean value.
+            If None is supplied all surveys are marked as being selected, if
+            a list of IDs (in form of ints) is supplied, only those surveys are
+            marked with True that exist in the DB and have a matching ID to one
+            of the given IDs.
+
+        Returns
+        -------
+        list of (int, str, bool)
```
josenavas: Why is the last boolean needed? It is also confusing that the selected attribute is only used to mark this boolean - shouldn't it limit which ones are returned, rather than the user then going back over the list to filter the ones that he is not interested in? And by user I mean the developer that consumes this function.

knimin/lib/mem_zip.py
```diff
@@ -61,6 +61,42 @@ def write_to_buffer(self):
         return self.in_memory_data.getvalue()
 
 
+def extract_zip(input_zip):
```
josenavas: Can this function blow up the memory usage?

knimin/lib/mem_zip.py
```diff
@@ -61,6 +61,42 @@ def write_to_buffer(self):
...
+    input_zip = zipfile.ZipFile(input_zip)
+    return {name: input_zip.read(name) for name in input_zip.namelist()}
+
+
+def sneak_files(archive, len=1000):
+    """ Returns the first characters of each file in an zip archive.
+
+    Parameters
+    ----------
+    len : int
```
josenavas: `archive` is missing

knimin/lib/tests/test_data_access.py
```diff
@@ -329,6 +328,23 @@ def test_get_ag_barcode_details(self):
             self.assertEqual({k: obs[key][k] for k in exp[key]}, exp[key])
             self.assertIn(obs[key]['participant_name'], participant_names)
 
+    def test_list_ag_surveys(self):
+        truth = [(-1, 'Personal Information', True),
```
josenavas: I think this test can be simplified as
```python
self.assertItemsEqual(db.list_ag_surveys(), truth)
```

knimin/templates/ag_pulldown.html
```diff
@@ -71,14 +80,34 @@ <h3 style="color:red">Pulldown Processing, please wait for file download. It may
 <form enctype="multipart/form-data" action="/ag_pulldown/" name="agForm" id="agForm" method="post">
 <p>Upload qiime mapping file, or other file with first line header and one barcode per line</p>
 <p>Barcodes File <input type="file" name="barcodes" id="barcodes"/></p>
+<p>American Gut Surveys:
+  <div>Select which American Gut surveys shall be included in the pulldown archive. <br/>Add 'all in one file' if you also want to have one addition file<br/>that contains all available columns from all American Gut surveys.</div>
```
josenavas: addition -> additional

knimin/tests/test_ag_pulldown.py
```diff
@@ -53,6 +55,13 @@ def test_get(self):
             self.assertIn("<option value='%s'>%s</option>" % (survey, survey),
                           response.body)
         self.assertNotIn('<input type="submit" disabled>', response.body)
+        for (id, name, selected) in db.list_ag_surveys():
```
josenavas: I think in this case none are selected, correct? according to the `get` call that you are doing all should be false.

knimin/handlers/ag_pulldown.py
sjanssen2: fixed

knimin/handlers/ag_pulldown.py
sjanssen2: good catch. Fixedgood catch. Fixed

knimin/lib/data_access.py
sjanssen2: better now?better now?

knimin/lib/data_access.py
sjanssen2: This function is used to render all survey names on the metadata pulldown website. The user (who is browsing labadmin) has the option to tick a survey to include or exclude it for the actual pulldown.
Thus, this function is a hybrid of DB query and interface rendering, but since it is so simple I thought it would be OK.

knimin/lib/data_access.py
sjanssen2: fixed

knimin/lib/mem_zip.py
sjanssen2: for sure, if you extract a huge archive. But it need to do the extraction for unit testing and I thought it is more elegant than doing system calls to 'unzip'

knimin/lib/tests/test_data_access.py
sjanssen2: nope. Consider that we add surveys in the future. This would brake your test.

knimin/templates/ag_pulldown.html
sjanssen2: fixed

knimin/tests/test_ag_pulldown.py
sjanssen2: remember the definition of list_ag_surveys. There it is said that all get "selected" if selected is not set, i.e it is None. So in fact, all are selected in this case.

sjanssen2:
ready for another round of reviews :-)

sjanssen2:
@josenavas could you review again please?

sjanssen2:
@wasade @antgonza @josenavas I need some timely reviews please!

knimin/handlers/ag_pulldown.py
```diff
@@ -55,10 +63,19 @@ def post(self):
             blanks = self.get_argument('blanks').split(',')
         else:
             blanks = []
-        if self.get_argument('external'):
+
+        # query which surveys have been selected by the user
+        if self.get_argument('selected_ag_surveys', []):
```
antgonza: is this if really necessary? you could do:
selected_ag_surveys = map(int, self.get_argument('selected_ag_surveys', []).split(','))
right?

knimin/handlers/ag_pulldown.py
```diff
@@ -55,10 +63,19 @@ def post(self):
...
+            selected_ag_surveys = map(int, self.get_argument(
+                'selected_ag_surveys').split(','))
+        else:
+            selected_ag_surveys = []
+
+        if self.get_argument('external', []):
```
antgonza: Same with this.

knimin/handlers/ag_pulldown.py
```diff
@@ -67,8 +84,43 @@ def post(self):
         failtext = ("The following barcodes were not retrieved "
                     "for any survey:\n%s" % failed)
         meta_zip.append("failures.txt", failtext)
+
+        # check database about what surveys are available
+        available_agsurveys = {}
+        for (id, name, selected) in db.list_ag_surveys():
```
antgonza: could you replace id -> _id and perhaps selected by _ as it's not being used?

knimin/lib/mem_zip.py
josenavas: ok!

knimin/lib/tests/test_data_access.py
josenavas: The test would not be correct, since it will not be testing that all surveys are returned.

knimin/lib/mem_zip.py
```diff
@@ -61,6 +61,45 @@ def write_to_buffer(self):
...
+        filenames and values are the content of the file.
+    len : int
+        Number of characters returned from each file.
+        Default: 1000.
+
+    Returns
+    -------
+    dict{str, str} where the first component is the filename and the second
+    the first <len> characters of the file."""
+    return map(lambda (k, v): {k: v[:len]}, archive.items())
```
antgonza: Just FYI, if the string is shorter it will return the full string ... this can cause problems in certain cases.

knimin/tests/test_ag_pulldown.py
josenavas: ok, in that case what it means is that you don't need the if statement because all are selected. Given the get issued, all should be selected so the test should actually be:
```python
for (_id, name, selected) in db.list_ag_surveys():
    self.assertIn("<option value='%i' selected>%s</option>" % (_id, name), response.body)
```

Note also the change from `id` -> `_id`.

knimin/handlers/ag_pulldown.py
sjanssen2: I figured that it can be even easier than that, if we use get_aguments (not the s). Then, you can provide a list of strings as parameters.

knimin/handlers/ag_pulldown.py
sjanssen2: see above

knimin/handlers/ag_pulldown.py
sjanssen2: done

knimin/lib/mem_zip.py
sjanssen2: should be fine in this case, but thanks for the info

knimin/handlers/ag_pulldown.py
sjanssen2: also catched by Antonio. Fixed.

knimin/handlers/ag_pulldown.py
```diff
@@ -67,8 +84,43 @@ def post(self):
...
+            # Currently, I (Stefan Janssen) don't have test data for external
+            # surveys, thus I don't know their 'survey' value. I expect it to
+            # be the name of the external survey. In order to not block their
+            # pulldown I check that a skipped survey ID must be in the set of
+            # all available surveys.
+            abs_survey = abs(survey)
+            if (abs_survey in selected_ag_surveys) or \
+               (abs_survey not in available_agsurveys):
+                meta_zip.append('survey_%s_md.txt' %
+                                available_agsurveys[-1 * survey], meta)
```
sjanssen2: that would be wrong, because those keys are actually negative numbers; it's what the user provided (i.e. the database)

knimin/lib/mem_zip.py
sjanssen2: I now know :-)

knimin/handlers/ag_pulldown.py
josenavas: Shouldn't it then be just `survey`? Note that survey holds the negative number, while `abs_survey` holds the positive one. `[-1 * survey] == abs_survey` as the code stands.

sjanssen2:
@josenavas regarding test_list_ag_surveys:
One of @mortonjt first comments addressed that and I changed my originally static tests to account for future addition of surveys. You know ask me to revert back. I don't how to satisfy both of you. Personally, I prefer Jamies way.

josenavas:
@sjanssen2 Unit tests should test that the functions are doing what they are supposed to do. The current get call on test_list_ag_surveys is used to test that all surveys are returned. If we add another survey and there is a bug in which we don't return the new survey, the test will pass w/o detecting the issue.

However, if we change to use assertItemsEqual, if a new survey is added the test would fail, which is the expected behavior since you will also need to update the test to make sure that the function still does what it is supposed to do, correct?

knimin/handlers/ag_pulldown.py
sjanssen2: I think I now fixed it. We cannot change the keys in available_agsurveys which are negative, but we can flip the sign for variable survey :-)

sjanssen2:
@josenavas regarding test_list_ag_surveys:
OK, reverted as you said.

antgonza:
Just remembered that you can retrieve a list with: self.request.arguments.get('yourvar', your_default_value)

sjanssen2:
@antgonza unfortunately, its not working here :-(

antgonza:
k, thanks for testing.

sjanssen2:
ready to merge @josenavas ?

sjanssen2:
would be great to get this done before I leave into the weekend, which is in 20 minutes

sjanssen2:
finally. Cool! Thanks @josenavas @antgonza @mortonjt

PR #167, opened 2016-12-20
updated expectation for one geocoder test

sjanssen2:
I think the test is broken. The expected location is a street in the Swedish city Eskilstuna. The actual street name is "Erlangengatan 12" not "12 Erlangen". However, there is a German city called "Erlangen". I guess that google updated it's approximate search such that it now gives precedence to a matching city name, rather to a misspelled Street name.

Thus, I tend to say that we have to correct our expectation in the test. Could you make this change? I suggest to replace '12 Erlangen' by 'Erlangengatan 12'.


wasade:
Thanks! :+1: pending internal email discussion

PR #172, opened 2017-03-22
proper insertion and roll back for third party survey tests

sjanssen2:
I re-designed the assumptions made for running unit tests for external survey sources, i.e. I carefully ensured that necessary data are inserted and reverted for those tests that need it. I chose not to use setup() and teardown() because those tests are doing this at some points implicitly.

@josenavas @mortonjt @wasade please give this a quick review to avoid merge conflicts with other people how are working on those tests!


sjanssen2:
coverage went down because of a new function for unit testing in data_access, which I cannot unit test itself

sjanssen2:
That's a cool idea @josenavas 
I have integrated it!

knimin/tests/test_ag_third_party.py
```diff
@@ -9,19 +9,21 @@
...
-    def setUp(self):
-        # Make sure vioscreen survey exists in DB
-        try:
-            db.add_external_survey('Vioscreen', 'FFQ', 'http://vioscreen.com')
-        except ValueError:
-            pass
-        super(AGThirdPartyHandler, self).setUp()
+    def tearDown(self):
+        for f in self._clean_up_funcs:
+            f()
```
josenavas: Awesome! I just realized that depending on where the test failed this call may also fail - probably safer adding a try/except around in which we just print on the screen that the clean up failed? This way if there is more than one call on the clean up all of them get executed.

sjanssen2: I added a try catch block for f()

josenavas:
👍 @qiyunzhu can you do a quick review?

sjanssen2:
@qiyunzhu @mortonjt @wasade would be greate to get a review and merge within the next 15 minutes to be able to push the next fix

knimin/tests/test_ag_third_party.py
```diff
@@ -9,19 +9,25 @@
...
+                      'description': 'FFQ',
+                      'url': 'http://vioscreen.com'}
+    _clean_up_funcs = []
+
+    def tearDown(self):
+        for f in self._clean_up_funcs:
+            try:
+                f()
+            except:
+                print("clean-up failed for. Downstream tests might be affected"
```
antgonza: Mind expanding explanation? "clean-up failed for ??"

knimin/tests/test_ag_third_party.py
```diff
@@ -140,21 +164,29 @@ def test_post_wrong_arguments(self):
...
-            pass
-        super(AGNewThirdPartyHandler, self).setUp()
+    _clean_up_funcs = []
+
+    def tearDown(self):
+        for f in self._clean_up_funcs:
+            try:
+                f()
+            except:
+                print("clean-up failed for. Downstream tests might be affected"
```
antgonza: I guess it means something to the devs ... but not clear if you are new.

knimin/tests/test_ag_third_party.py
sjanssen2: It means that something has not been deleted in the DB which was inserted previously for executing unit tests.
I think also specifying what failed will be some overhead development and goes into the direction of the already existing teadDown() method.

sjanssen2:
is it good to merge now @josenavas ?

knimin/tests/test_ag_third_party.py
```diff
@@ -9,19 +9,25 @@
...
+                      'description': 'FFQ',
+                      'url': 'http://vioscreen.com'}
+    _clean_up_funcs = []
+
+    def tearDown(self):
+        for f in self._clean_up_funcs:
+            try:
+                f()
+            except:
+                print("Database clean-up failed. Downstream tests might be"
```
josenavas: @sjanssen2 I think what @antgonza meant here was to do something along these lines:
```python
try:
    f()
except Exception as e:
    print("Database clean-up failed. Downstream test might be affected by this! Reason: %s" % str(e))
```

sjanssen2: Oh I see, changed it as you suggested!

PR #173, opened 2017-03-22
Fix ut ag search

sjanssen2:
ensure that strings are properly xhtml escaped

### DIFF
--- a/knimin/handlers/ag_search.py
+++ b/knimin/handlers/ag_search.py
@@ -1,9 +1,11 @@
 #!/usr/bin/env python
 from tornado.web import authenticated
+from tornado.escape import xhtml_escape
 from knimin.handlers.base import BaseHandler
 from knimin.handlers.access_decorators import set_access
 
 from knimin import db
+from knimin.lib.util import xhtml_escape_recursive
 
 
 @set_access(['Search'])
@@ -58,5 +60,7 @@ class AGSearchHandler(BaseHandler):
             display_results.append(login_display)
 
         # now render the page
-        self.render("ag_search.html", results=display_results,
-                    handouts=handouts, currentuser=self.current_user)
+        self.render("ag_search.html",
+                    results=xhtml_escape_recursive(display_results),
+                    handouts=map(xhtml_escape, handouts),
+                    currentuser=self.current_user)
--- a/knimin/handlers/ag_search.py
+++ b/knimin/handlers/ag_search.py
@@ -1,11 +1,9 @@
 #!/usr/bin/env python
 from tornado.web import authenticated
-from tornado.escape import xhtml_escape
 from knimin.handlers.base import BaseHandler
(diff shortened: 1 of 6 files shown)


sjanssen2:
please give me your thoughts about this PR @josenavas @antgonza

knimin/handlers/ag_search.py
```diff
@@ -58,5 +60,7 @@ def post(self):
             display_results.append(login_display)
 
         # now render the page
-        self.render("ag_search.html", results=display_results,
-                    handouts=handouts, currentuser=self.current_user)
+        self.render("ag_search.html",
+                    results=xhtml_escape_recursive(display_results),
+                    handouts=map(xhtml_escape, handouts),
```
josenavas: @sjanssen2 I'm wondering why the text should be escaped here? These values are used to render the template on the server side, so this ends up going to the user in the body of the html, which shouldn't need to explicitly escaped. Just commenting once - but applies to all calls to "render".

sjanssen2: because the text might contain utf8 values which - I think - are rendered wrong if not escaped. But I'd be happy if you could double check.

sjanssen2: cool. Looks like you are right and escaping is not necessary for rendering, but for unit testing.

knimin/lib/util.py
```diff
@@ -302,3 +303,23 @@ def categorize_bmi(x):
         bmi_cat = 'Unspecified'
 
     return bmi_cat
+
+
+def xhtml_escape_recursive(d):
```
josenavas: Can you add a test for this?

knimin/tests/test_ag_search.py
```diff
@@ -64,7 +65,9 @@ def test_post_barcode(self):
                              field == 'ag_kit_barcode_id'):
                             self.assertNotIn(sample[field], response.body)
                         else:
-                            self.assertIn(sample[field], response.body)
+                            exp = xhtml_escape_recursive(sample[field])
```
antgonza: I'm a bit confused about this test. Basically, you have kit_id, then from that kit_id you loop over all the fields to check that they are in the response right? However, this should be a deterministic output, right? Not sure why the need to all these loops ...

sjanssen2: because it depends on the values in the DB which are not guaranteed to stay the same when we switch to Jose's new dump

antgonza: If @josenavas is fine with this, then OK, cause IMOO this defeats the idea of deterministic tests and basically if some code changes something in the DB, the test will not catch it; but perhaps this is fine for this projects ...

knimin/lib/util.py
sjanssen2: ups, I forgot. It's now in there.

josenavas:
@sjanssen2 1 flake8 error

PR #174, opened 2017-03-23
fix: utf8 zipcodes

sjanssen2:
fixed problems regarding utf-8 chars in zipcodes. Specifically, error messages need explicit encoding and so does SQL like and dict lookups


mortonjt:
Is it possible to add a test for this update?

josenavas:
@mortonjt the new database contains utf-8 characters, so once we merge [this PR](https://github.com/biocore/american-gut-web/pull/645) all these changes will be tested. Before merging the other PR, however, we are making sure that all test will pass here in labadmin, so @sjanssen2 and I are making the needed changes for this to happen.

Do you have any comment regarding the code changed here?

sjanssen2:
ready to merge?

PR #178, opened 2017-03-24
fixing barcode_utils tests

sjanssen2:
proper encoding of utf8 fields
getting barcodes from DB which are unassigned or handed out

### DIFF
--- a/knimin/lib/data_access.py
+++ b/knimin/lib/data_access.py
@@ -2362,13 +2362,14 @@ class KniminAccess(object):
         sql = """SELECT project from barcodes.project
                  JOIN barcodes.project_barcode USING (project_id)
                  where barcode = %s"""
-        results = [x[0] for x in self._con.execute_fetchall(sql, [barcode])]
-        if 'American Gut Project' in results:
-            parent_project = 'American Gut'
-            results.remove('American Gut Project')
-            projects = ', '.join(results)
+        results = [self._unicode_convert(x[0])
+                   for x in self._con.execute_fetchall(sql, [barcode])]
+        if u'American Gut Project' in results:
+            parent_project = u'American Gut'
+            results.remove(u'American Gut Project')
+            projects = u', '.join(results)
         else:
-            projects = ', '.join(results)
+            projects = u', '.join(results)
             parent_project = projects
         return (projects, parent_project)
 
@@ -2405,7 +2406,8 @@ class KniminAccess(object):
         """Returns a list of project names
         """
         sql = """SELECT project FROM project"""
-        return [x[0] for x in self._con.execute_fetchall(sql)]
+        return [self._unicode_convert(x[0])
+                for x in self._con.execute_fetchall(sql)]
 
     def set_deposited_ebi(self):
         """Updates barcode deposited status by checking EBI"""
@@ -2575,3 +2577,23 @@ class KniminAccess(object):
         if not info:
             raise ValueError('No unassigned barcodes found.')
         return info[0]
+
+    def ut_remove_project(self, project_name):
+        """ Deletes a project.
+        For unit testing only!
+
+        Parameters
+        ----------
+        project_name : str
+            The name of the project that should be deleted.
+
+        Raises
+        ------
+        ValueError
+            If project could not be deleted, e.g. because barcodes are still
+            associated to this project."""
+        sql = """DELETE FROM barcodes.project WHERE project=%s"""
+        try:
+            self._con.execute(sql, [project_name])
+        except:
+            raise ValueError("Unable to delete project %s" % project_name)
--- a/knimin/lib/data_access.py
+++ b/knimin/lib/data_access.py
@@ -2564,8 +2564,7 @@ class KniminAccess(object):
         return info[0]
 
     def ut_get_arbitrary_unassigned_barcode(self):
-        """ Returns an arbitrarily chosen barcode that is not yet assigned to
-        a survey.
+        """ Returns an arbitrarily chosen unassigned barcode.
         For unit testing only!
 
         Returns
--- a/knimin/lib/data_access.py
+++ b/knimin/lib/data_access.py
@@ -2440,6 +2440,7 @@ class KniminAccess(object):
 
     def ut_remove_external_survey(self, name, description, url):
         """ Remove an external survey from DB.
+        For unit testing only!
 
         Parameters
         ----------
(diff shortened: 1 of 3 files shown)


sjanssen2:
WAIT, I tested against the wrong DB version. Thus, this is work in progres. Don't merge!

josenavas:
@sjanssen2 If I understand correctly, this is the last PR needed to change all the tests to use the new scrubbed DB. Do you need any help trying to finish these ones?

sjanssen2:
Ready for reviews @antgonza @josenavas @wasade

knimin/lib/data_access.py
```diff
@@ -2523,3 +2525,77 @@ def ut_get_supplied_kit_id(self, ag_login_id):
             raise ValueError('ag_login_id not in database: %s' %
                              ag_login_id)
         return info[0][0]
+
+    def ut_get_arbitrary_handout_barcode(self, is_AGP=True):
+        """ Returns an arbitrarily chosen barcode that is handed out.
+            If is_AGP is True, this barcode belongs to the AGP, otherwise to
```
antgonza: Should these lines be Note?

knimin/lib/data_access.py
```diff
@@ -2523,3 +2525,77 @@ def ut_get_supplied_kit_id(self, ag_login_id):
...
+        sql += """ LIMIT 1"""
+        info = self._con.execute_fetchone(sql, [])
+        if not info:
+            raise ValueError('No handout barcodes found.')
+        return info[0]
+
+    def ut_get_arbitrary_unassigned_barcode(self):
+        """ Returns an arbitrarily chosen barcode that not yet assigned to
+            a survey.
+            For unit testing only!
```
antgonza: Is this line necessary in all code? Perhaps ut_* should suffice, right?

knimin/tests/test_barcode_util.py
```diff
@@ -307,23 +318,27 @@ def test_get_ag_details_SJ(self):
...
+               'barcode': '000001018', 'moldy_checked': '',
+               'date_of_last_email': '', 'site_sampled': 'Stool',
+               # 'email_type': '1',
+               # 'name': 'REMOVED',
+               'sample_time': time(6, 50),
+               # 'notes': 'REMOVED',
+               # 'email': 'REMOVED'
+              }
+        # only look at those fields, that are not subject to scrubbing
+        self.assertEqual({k: ag_details[k] for k in exp}, exp)
```
antgonza: Could you add a check for the keys? In case, that in the future we add more elements, we know that we are actually testing those ...

knimin/lib/data_access.py
```diff
@@ -2523,3 +2525,77 @@ def ut_get_supplied_kit_id(self, ag_login_id):
...
+        if is_AGP:
+            sql += """ WHERE project_id=1"""
+        sql += """ LIMIT 1"""
+        info = self._con.execute_fetchone(sql, [])
+        if not info:
+            raise ValueError('No handout barcodes found.')
+        return info[0]
+
+    def ut_get_arbitrary_unassigned_barcode(self):
+        """ Returns an arbitrarily chosen barcode that not yet assigned to
```
wasade: I believe this method is actually getting an unassigned barcode *irrespective of project*

knimin/lib/data_access.py
```diff
@@ -2523,3 +2525,77 @@ def ut_get_supplied_kit_id(self, ag_login_id):
...
+            Returns
+            -------
+            str: barcode
+                Example: '000001000'
+
+            Raises
+            ------
+            ValueError
+                If no unassigned barcode can be found in the DB.
+            """
```
wasade: it looks like the docstring is overindented

knimin/lib/data_access.py
sjanssen2: have you seen these two lines:
`if is_AGP:
            sql += """ WHERE project_id=1"""
        `

knimin/lib/data_access.py
sjanssen2: indentation is now corrected

knimin/lib/data_access.py
sjanssen2: yes. I changed it.

knimin/lib/data_access.py
sjanssen2: I would like to leave it as is, because there are so many locations where I would have to change it, also in american-gut-web

knimin/tests/test_barcode_util.py
josenavas: @sjanssen2 I think you're missing fixing this comment.

knimin/lib/data_access.py
wasade: Those lines aren't in this function...

knimin/tests/test_barcode_util.py
sjanssen2: I added it now

knimin/lib/data_access.py
sjanssen2: sorry, I was looking at ut_get_arbitrary_handout_barcode. My fault.
Yes, you are right, it returns one barcode not assigned to any project. But I don't see why you made this comment :-/

josenavas:
@wasade can you check if your comments have been addressed?

knimin/lib/data_access.py
```diff
@@ -2523,3 +2526,80 @@ def ut_get_supplied_kit_id(self, ag_login_id):
...
+            sql += """ WHERE project_id=1"""
+        sql += """ LIMIT 1"""
+        info = self._con.execute_fetchone(sql, [])
+        if not info:
+            raise ValueError('No handout barcodes found.')
+        return info[0]
+
+    def ut_get_arbitrary_unassigned_barcode(self):
+        """ Returns an arbitrarily chosen barcode that is not yet assigned to
+        a survey.
```
wasade: The docstring says "not yet assigned to a survey" but if this is an arbitrary barcode irrespective of project, then there may not be a survey to associate too

sjanssen2: I am getting confused. Could you please write down how you want to have it changed?

wasade: "Returns an arbitrarily chosen barcode that is not yet assigned to a survey."

to

"Returns an arbitrarily chosen unassigned barcode."

josenavas:
Thanks @wasade and @sjanssen2 !! @wasade wanna merge?

josenavas:
Merging since @wasade's comments have been addressed

PR #179, opened 2017-03-24
fixing knimin/tests/test_ag_new_barcode.py

josenavas:
@sjanssen2 @antgonza able for a quick review?

### DIFF
--- a/knimin/tests/test_ag_new_barcode.py
+++ b/knimin/tests/test_ag_new_barcode.py
@@ -21,7 +21,7 @@ class TestAGBarcodePrintoutHandler(TestHandlerBase):
         self.assertEqual(response.headers['Content-Disposition'],
                          'attachment; filename=barcodes.pdf')
         # check that the files is a PDF ...
-        self.assertIn('%PDF-1.4', response.body)
+        self.assertIn('%PDF-1.', response.body)
         # ... and it is not empty
         self.assertEqual(len(response.body) > 1000, True)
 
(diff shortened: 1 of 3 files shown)


knimin/tests/test_ag_new_barcode.py
```diff
@@ -69,7 +69,7 @@ def test_post(self):
         self.mock_login_admin()
         action = 'unknown'
         num_barcodes = 4
-        projects = ["American Gut Project", "Autism Spectrum Disorder"]
+        projects = ["American Gut Project"]
```
sjanssen2: I'd recommend to test against a utf8 encoded project name.

knimin/tests/test_ag_new_barcode.py
```diff
@@ -21,7 +21,7 @@ def test_post(self):
         self.assertEqual(response.headers['Content-Disposition'],
                          'attachment; filename=barcodes.pdf')
         # check that the files is a PDF ...
-        self.assertIn('%PDF-1.4', response.body)
+        self.assertIn('%PDF-1.', response.body)
```
antgonza: Why this change? Making sure that any 1* version works?

josenavas: Yes, in my machine it was failing because my PDF is 1.5

josenavas:
There was some stochasticity in one of the tests that I just fixed - @antgonza do you want to make the honors?

PR #180, opened 2017-03-24
Fixing knimin/tests/test_ag_update_geocode.py

josenavas:
I had to add a limit on the test cause otherwise it will try to geocode the entire DB, which will take a while.

@sjanssen2 @antgonza quick review?


sjanssen2:
looks good to me, lets wait for the test results

PR #181, opened 2017-03-24
Adding jira

antgonza:
This is an initial test to see if we can add jira to the labadmin. Note that the current code is connecting to the test site of jira: https://jira.atlassian.com, as an anonymous user.


wasade:
:+1:

sjanssen2:
don't we have a test jira instance?

josenavas:
I marked it as a WIP PR. We currently don't have a test Jira instance.

knimin/lib/ag_jira.py
```diff
@@ -0,0 +1,42 @@
...
+from jira import JIRA
+
+
+def jira_connect(server='https://jira.atlassian.com',
+                 user=None, password=None):
+    """Connect to JIRA server
+
+    Parameters
+    ----------
+    server: str, optional
```
sjanssen2: why is server optional? Can it work without specifying a server? I would vote for making server mandatory

antgonza: It's optional cause it has a default value, which is valid. BTW optional in this context means that the user doesn't have to pass it ...

.travis.yml
```diff
@@ -6,12 +6,26 @@ services:
   - redis-server
 addons:
   postgresql: "9.3"
+  packages:
+    - oracle-java8-installer
```
wasade: :(

knimin/lib/tests/test_ag_jira.py
```diff
@@ -0,0 +1,19 @@
+from unittest import TestCase, main
+
+from knimin.lib.ag_jira import jira_connect, get_projects
```
wasade: should this be `ag_jira` or just `jira`?

PR #185, opened 2017-03-29
Enforcing tornado 4.2.2

josenavas:


wasade:
I think travis died, restarted it

josenavas:
@sjanssen2 you can merge given that it is a really simple change.

PR #187, opened 2017-04-03
Create plate

josenavas:
Not sure why there are so many commits in here, my branch is already in sync with upstream/plate-mapper-kiwi. Anyways, the diff of the files is correct and shows the changes that I've made.

Here I add the first page of the plate mapper, used to create a plate. See below for gif of the page.
Few comments about this PR:
- In the GIF, after submission it shows a 404 page. This is expected as the next page is still not in place (next PR).
- The PR seems big, but I've added a bunch of JavaScript libraries (bootstrap, select2, dropdown), their CSS files and their licences, so the actual PR is smaller.
- In the file `test_data_access.py` I've commended out some tests. Those tests are currently failing, but I don't want to fix them at this moment as those functions will change on my next PR.

sample_plate_creation


antgonza:
From the text and the tests commented out it sounds like tests should be passing, right? Could you fix them?

josenavas:
Thanks @antgonza I fixed one of them and commented out the other one as I also need to modify that page later.

knimin/handlers/ag_search.py
```diff
@@ -49,11 +49,8 @@ def post(self):
                     barcode_info[ag_barcode['barcode']]['ag_info'] = ag_barcode
                     lab_barcode_info = db.get_barcode_details(
                         ag_barcode['barcode'])
-                    plate = db.get_plate_for_barcode(
```
antgonza: Why the delete? Are we not retrieving the plate here anymore?

knimin/lib/data_access.py
```diff
@@ -2469,88 +2482,83 @@ def _study_is_unique(self, qiita_study_id=None, title=None, skip_id=None):
...
-            err_msg = "%s %s conflicts with exisiting study %s."
-            for col, val, tag in zip(cols, vals, tags):
-                if val:
-                    TRN.add(sql.format(col), [val, skip_id])
-                    res = TRN.execute_fetchflatten()
-                    if res:
-                        errs.append(err_msg % (tag, repr(val), res[0]))
-            if errs:
-                raise ValueError('\n'.join(errs))
+                     WHERE (study_id = %s OR title = %s)"""
```
antgonza: Note that this will make titles case sensitive, so we can have "My Title" and "my Title". Perhaps fine but wanted to raise it.

knimin/lib/data_access.py
```diff
@@ -2469,88 +2482,83 @@ def _study_is_unique(self, qiita_study_id=None, title=None, skip_id=None):
...
+            if skip_id:
+                sql += " AND study_id != %s"
+                sql_args.append(skip_id)
 
-    def create_study(self, qiita_study_id=None, title=None, alias=None,
-                     notes=None):
+            TRN.add(sql, sql_args)
+            res = TRN.execute_fetchflatten()
+            if res:
+                raise ValueError("Study (%s, %s) conflicts with studies %s"
```
antgonza: This function is not tested directly, AFAIK, and this error is also not tested. In fact, I _think_ this error will not work as is and should have: `', '.join(map(str, res[0]))))` ...

knimin/lib/data_access.py
```diff
@@ -3015,8 +3044,7 @@ def _plate_type_exists(self, id):
             if not TRN.execute_fetchlast():
                 raise ValueError('Plate type ID %s does not exist.' % id)
 
-    def create_sample_plate(self, name, plate_type_id, email=None,
-                            created_on=None, notes=None):
+    def create_sample_plate(self, name, plate_type_id, email, studies):
```
antgonza: No test in the comments or outside the comments.

knimin/lib/data_access.py
```diff
@@ -2930,6 +2938,27 @@ def _sample_plate_exists(self, id):
             if not TRN.execute_fetchlast():
                 raise ValueError('Sample plate ID %s does not exist.' % id)
 
+    def sample_plate_name_exists(self, name):
```
antgonza: No tests in the comments or outside the comments.

knimin/handlers/ag_search.py
josenavas: This call was accessing to an old plate structure that no longer exists. That plate structure was always empty since the plate mapper was never in place. Apparently, during the first iteration of the plate mapper (quite a while back) some DB changed made it to master but they where never populated.

knimin/lib/data_access.py
josenavas: Good point - I went and check on what [Qiita does](https://github.com/biocore/qiita/blob/master/qiita_db/study.py#L284) (which also requires the study title to be unique) and I noticed that the study title is case sensitive.

I think we should do here whatever Qiita does.

knimin/lib/data_access.py
josenavas: Tested here: https://github.com/biocore/labadmin/pull/187/files#diff-0d1d77e117ce5a979751aa7fba90e5e5R377

Note that is `fetchflatten` so res is a list of ints.

knimin/lib/data_access.py
josenavas: Tested [here](https://github.com/biocore/labadmin/pull/187/files#diff-0d1d77e117ce5a979751aa7fba90e5e5R684)

knimin/lib/data_access.py
josenavas: Tested [here](https://github.com/biocore/labadmin/pull/187/files#diff-0d1d77e117ce5a979751aa7fba90e5e5R699)

knimin/lib/data_access.py
```diff
@@ -3083,28 +3114,31 @@ def edit_sample_plate(self, id, name, plate_type_id, email=None,
...
         with TRN:
-            self._sample_plate_exists(id)
-            sql = """SELECT name, plate_type_id, email, created_on, notes
+            self._sample_plate_exists(plate_id)
+            sql = """SELECT name, plate_type_id, email, created_on, notes,
+                        array_agg(study_id ORDER BY study_id) as studies
                      FROM pm.sample_plate
-                     WHERE sample_plate_id = %s"""
-            TRN.add(sql, [id])
+                        JOIN pm.sample_plate_study USING (sample_plate_id)
```
qiyunzhu: Is this `sample_plate_study` a table of sample plate ID to study ID maps?

knimin/lib/data_access.py
```diff
@@ -3198,21 +3232,24 @@ def read_sample_plate_layout(self, id):
...
             ID of the sample plate to delete
         """
         with TRN:
-            self._sample_plate_exists(id)
-            if self._sample_plate_layout_exists(id):
-                self._clear_sample_plate_layout(id)
+            self._sample_plate_exists(plate_id)
+            if self._sample_plate_layout_exists(plate_id):
+                self._clear_sample_plate_layout(plate_id)
+            sql = """DELETE fROM pm.sample_plate_study
```
qiyunzhu: Is it possible that a created sample plate has no study associated?

knimin/templates/pm_create_plate.html
```diff
@@ -0,0 +1,119 @@
...
+    if (studies.length === 0) {
+      // The user didn't choose any study
+      $('#studies-warning').show();
+    }
+    if (plateName.length === 0) {
+      // The plate name is empty or is just spaces
+      $('#plate-name-warning').html('* Please introduce a plate name');
+      $('#plate-name-warning').show();
+    } else {
+      $.get('/pm_sample_plate/name_check?name=' + plateName, function(data){
```
qiyunzhu: Is it possible that `plateName` contains special characters that cause escape issues? Also better have a space between `)` and `{`.

knimin/handlers/pm_plate_map.py
```diff
@@ -0,0 +1,42 @@
...
+        self.redirect('/pm_plate_map/%d/' % plate_id)
+
+
+@set_access(['Admin'])
+class PMPlateNameCheckerHandler(BaseHandler):
+    @authenticated
+    def get(self):
+        name = self.get_argument('name')
+        self.write({'result': db.sample_plate_name_exists(name)})
+        self.finish()
```
wasade: should this return a status code 404 if a plate does not exist?

knimin/templates/pm_create_plate.html
```diff
@@ -0,0 +1,119 @@
...
+        .removeClass('btn-add').addClass('btn-remove')
+        .removeClass('btn-success').addClass('btn-info')
+        .html('<span class="glyphicon glyphicon-minus"></span>');
+    }).on('click', '.btn-remove', function(e) {
+        $(this).parents('.entry:first').remove();
+        e.preventDefault();
+        return false;
+    });
+
+    // Something
```
qiyunzhu: Is there something incomplete here?

knimin/lib/data_access.py
```diff
@@ -2469,88 +2482,83 @@ def _study_is_unique(self, qiita_study_id=None, title=None, skip_id=None):
...
-                if val:
-                    TRN.add(sql.format(col), [val, skip_id])
-                    res = TRN.execute_fetchflatten()
-                    if res:
-                        errs.append(err_msg % (tag, repr(val), res[0]))
-            if errs:
-                raise ValueError('\n'.join(errs))
+                     WHERE (study_id = %s OR title = %s)"""
+            sql_args = [qiita_study_id, title]
+            if skip_id:
```
wasade: what's the use case for this?

knimin/templates/pm_create_plate.html
```diff
@@ -0,0 +1,119 @@
...
+    <button class="btn btn-success btn-add" type="button"><span class="glyphicon glyphicon-plus"></span></button>
+  </div>
+</div>
+<label>Plate name: <b id='plate-name-warning' class='warning' hidden></b></label>
+<input class='form-control' id='plate-name'>
+</br>
+<div class="form-group">
+  <button onclick="submit_plate_creation();" class="btn button-sm btn-primary">Create plate</button>
+</div>
+</div>
```
qiyunzhu: Is this `</div>` redundant?

knimin/lib/data_access.py
wasade: fROM -> FROM

knimin/lib/tests/test_data_access.py
```diff
@@ -487,409 +562,428 @@ def test_create_samples(self):
...
+    #     obs = db.get_samples_by_study(sids[2])
+    #     self.assertDictEqual(obs, {})
+    #     # Attempt to read a non-existing study
+    #     db.delete_samples(['1', '2'])
+    #     for sid in sids:
+    #         db.delete_study(sid)
+    #     with self.assertRaises(ValueError) as context:
+    #         db.get_samples_by_study(sids[0])
+    #     err = 'Study ID %s does not exist.' % sids[0]
+    #     self.assertEqual(str(context.exception), err)
```
mortonjt: A lot of commented code.  This is necessary anymore?

knimin/lib/tests/test_data_access.py
```diff
@@ -487,409 +562,428 @@ def test_create_samples(self):
...
+    #         db.edit_sample_plate(spid, 'test_plate', ptid,
+    #                              email='not-an-email')
+    #     err = 'Email not-an-email does not exist.'
+    #     self.assertEqual(str(context.exception), err)
+    #     # Attempt to edit a sample plate that does not exist
+    #     db.delete_sample_plate(spid)
+    #     with self.assertRaises(ValueError) as context:
+    #         db.edit_sample_plate(spid, **spinfo)
+    #     err = 'Sample plate ID %s does not exist.' % spid
+    #     self.assertEqual(str(context.exception), err)
```
mortonjt: Same here

knimin/lib/tests/test_data_access.py
```diff
@@ -487,409 +562,428 @@ def test_create_samples(self):
...
+    #     obs = db.read_sample_plate_layout(spid)
+    #     self.assertListEqual(obs, [])
+    #     # Attempt to read layout of a non-existing sample plate
+    #     db.delete_sample_plate(spid)
+    #     with self.assertRaises(ValueError) as context:
+    #         db.write_sample_plate_layout(spid, splayout)
+    #     err = 'Sample plate ID %s does not exist.' % spid
+    #     self.assertEqual(str(context.exception), err)
+    #     db.delete_samples([x['id'] for x in samples])
+    #     db.delete_study(sid)
```
mortonjt: Same here

knimin/lib/tests/test_data_access.py
```diff
@@ -487,409 +562,428 @@ def test_create_samples(self):
...
+        # self.assertTrue(obs)
+        # db.delete_sample_plate(spid)
+        # with self.assertRaises(ValueError) as context:
+        #     db.read_sample_plate(spid)
+        # err = 'Sample plate ID %s does not exist.' % spid
+        # self.assertEqual(str(context.exception), err)
+        # obs = db._sample_plate_layout_exists(spid)
+        # self.assertFalse(obs)
+        # db.delete_samples(['1'])
+        # db.delete_study(sid)
```
mortonjt: Same here

knimin/tests/test_pm_plate_list.py
```diff
@@ -1,8 +1,8 @@
 from unittest import main
 
 from knimin.tests.tornado_test_base import TestHandlerBase
-from knimin import db
-import datetime
+# from knimin import db
+# import datetime
```
mortonjt: delete?

knimin/tests/test_pm_plate_list.py
```diff
@@ -13,74 +13,75 @@ def test_get_not_authed(self):
...
+    #                '\n'
+    #                '</span></td>\n'
+    #                '<td title="' + p1['study'][3] + '">' +
+    #                str(p1['study'][1]) + '</td>\n'
+    #                '<td>' + (p1['person'] or '-') + '</td>\n'
+    #                '<td>' + (p1['date'] or '-') + '</td>\n')
+    #         self.assertIn(exp, response.body)
+    #         db.delete_sample_plate(spid)
+    #     db.delete_samples(['test_sample_' + str(x) for x in range(total)])
+    #     db.delete_study(sid)
```
mortonjt: Same here

knimin/handlers/pm_plate_map.py
josenavas: Good call changed - I've added also some tests to these handlers - I noticed they're missing

knimin/lib/data_access.py
josenavas: In case that you're editing a study, if the new title is the same as the old is ok, so you can pass the id of the current study that you're editing to avoid failing because your title is the same as your own title. This is used to avoid code duplication and being able to use this function in a couple of places easily.

knimin/lib/data_access.py
```diff
@@ -2469,88 +2482,83 @@ def _study_is_unique(self, qiita_study_id=None, title=None, skip_id=None):
...
+
         with TRN:
             self._study_exists(study_id)
-            self._study_is_unique(qiita_study_id, title, study_id)
+            self._study_is_unique(study_id, title, study_id)
+            cols = []
+            sql_args = []
+            if title is not None:
+                cols.append("title = %s")
+                sql_args.append(title)
```
josenavas: These lines are creating the list of columns to be modified and the parameters passed. Note that `cols` is used in line 2561 in the `join` call, and `sql_args` is used when adding the call to the transaction object in line 2562.

knimin/lib/data_access.py
```diff
@@ -3040,15 +3066,20 @@ def create_sample_plate(self, name, plate_type_id, email=None,
...
             self._sample_plate_is_unique(name)
             self._plate_type_exists(plate_type_id)
-            if email:
-                self._email_exists(email)
+            self._email_exists(email)
             sql = """INSERT INTO pm.sample_plate (name, plate_type_id, email,
-                                                  created_on, notes)
-                     VALUES (%s, %s, %s, %s, %s)
+                                                  created_on)
+                     VALUES (%s, %s, %s, now())
```
josenavas: I thought about it, but given that we are unsure on how to transfer the old data to the new DB structure I thought on simplifying the API as much as possible. When used through the interface this should be set automatically to the current time.

knimin/lib/data_access.py
josenavas: Yes, this table maps the sample plate to the studies being plated. This is needed so we can filter the list of samples to populate the plate.

knimin/lib/data_access.py
josenavas: No, and in case that it doesn't have any, this call will not fail.

knimin/lib/tests/test_data_access.py
josenavas: As described in the PR description (applies to all so commenting only once):
> In the file `test_data_access.py` I've commended out some tests. Those tests are currently failing, but I don't want to fix them at this moment as those functions will change on my next PR.

knimin/templates/pm_create_plate.html
josenavas: Added the space. I've tried multiple characters (such `/` `'` `~` `@` `) and I couldn't break it.

knimin/templates/pm_create_plate.html
josenavas: Yes, the comment! Good catch 😄

knimin/templates/pm_create_plate.html
josenavas: Nope, matches the `<div>` on line 95. I've updated the indentation to make it clear

knimin/lib/data_access.py
wasade: okay

qiyunzhu:
@josenavas Your edits and comments sound good to me! Ready to ship!

PR #190, opened 2017-04-03
How to install JIRA

antgonza:
... and fixing some minor errors.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -21,10 +21,36 @@ Clone the repository, and pip install::
    pip install -e .
 
 Copy the example config file to be visible for starting up a test database::
-   
+
    cp ./knimin/config.txt.example ./knimin/config.txt
 
-You should now be able to start the webserver::
+Installing JIRA
+~~~~~~~~~~~~~~~
+
+Note that these instructions should work for Mac or Linux and that
+you need to have JAVA 1.8 or higher. If you are in a Mac, you will need to
+[update](http://www.oracle.com/technetwork/java/javase/downloads/jdk8-downloads-2133151.html).
+If you updated to that package you could force using it by::
+
+    export JAVA_HOME='/Library/Java/JavaVirtualMachines/jdk1.8.0_121.jdk/Contents/Home'
+
+Now you can continue with downloading the latest [JIRA SDK](https://marketplace.atlassian.com/download/plugins/atlassian-plugin-sdk-tgz),
+uncompressing it and setting it up::
+
+    curl -o atlassian-plugin-sdk-6.2.14.tar.gz https://marketplace.atlassian.com/download/plugins/atlassian-plugin-sdk-tgz
+    tar zxvf atlassian-plugin-sdk-6.2.14.tar.gz
+    mv atlassian-plugin-sdk-6.2.14 atlassian-plugin-sdk
+    export PATH="$PATH:${PWD}/atlassian-plugin-sdk/bin"
+
+To test the install you can run::
+
+    atlas-version
+
+and to start the JIRA system locally::
+
+    atlas-run-standalone --product jira </dev/zero 2>&1 &
+
+Now, you should now be able to start the webserver::
 
    python ./knimin/webserver.py
 
@@ -34,7 +60,7 @@ Initial default test login credentials are:
 
 **User:** test
 
-**Password:** password 
+**Password:** password
 
 Testing
 -------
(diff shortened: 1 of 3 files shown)


README.rst
```diff
@@ -21,10 +21,35 @@ Clone the repository, and pip install::
...
 
 Copy the example config file to be visible for starting up a test database::
-   
+
    cp ./knimin/config.txt.example ./knimin/config.txt
 
-You should now be able to start the webserver::
+**Installing JIRA**
+
+Note that these instructions should work for Mac or Linux and that
```
sjanssen2: what about using conda for this? conda install -c bioconda java-jdk

README.rst
```diff
@@ -21,10 +21,35 @@ Clone the repository, and pip install::
...
+If you updated to that package you could force using it by::
+
+    export JAVA_HOME='/Library/Java/JavaVirtualMachines/jdk1.8.0_121.jdk/Contents/Home'
+
+Now you can continue with downloading the latest `JIRA SDK <https://marketplace.atlassian.com/download/plugins/atlassian-plugin-sdk-tgz>`__,
+uncompressing it and setting it up::
+
+    curl -o atlassian-plugin-sdk-6.2.14.tar.gz https://marketplace.atlassian.com/download/plugins/atlassian-plugin-sdk-tgz
+    tar zxvf atlassian-plugin-sdk-6.2.14.tar.gz
+    mv atlassian-plugin-sdk-6.2.14 atlassian-plugin-sdk
```
sjanssen2: why are you doing that? Better have softlinks to be able to switch between different versions later

README.rst
antgonza: Good option. However, I don't have/use conda and these instructions work with or without it ...

README.rst
antgonza: Whatever works for you, this is one of many ways to do it. I'm simply documenting one of them ... Once this is merge you could add any others that you prefer.

README.rst
```diff
@@ -21,10 +21,35 @@ Clone the repository, and pip install::
...
+    export PATH="$PATH:${PWD}/atlassian-plugin-sdk/bin"
```
sjanssen2: aren't we conda guys and haven't we already set up an environment? So why not:
mkdir -p ~/miniconda3/envs/labadmin/etc/conda/activate.d
echo "export PATH=../../../src/atlassian-plugin-sdk-6.2.14/bin:$PATH" > setpath.sh

README.rst
```diff
@@ -21,10 +21,35 @@ Clone the repository, and pip install::
...
+
+To test the install you can run::
+
+    atlas-version
+
+and to start the JIRA system locally::
+
+    atlas-run-standalone --product jira </dev/zero 2>&1 &
```
sjanssen2: what is` </dev/zero` doing?

README.rst
antgonza: Like I said, I'm not, in fact I still virtualenv everything but if others feel like we need to change these lines I'm happy to do it ...

README.rst
antgonza: Good point, I documented all that in travis but not here, I can add here:
</dev/zero 2>&1 & --> https://goo.gl/n7BYnh

README.rst
sjanssen2: But we do ask the user to install conda and create an environment in the beginning of the readme. Therefore, I'd say lets use conda as the default and pro users like you can derive whatever is necessary to install in the root environment, i.e. without conda.
It would also be easier to remove the whole environment later on your machine.

README.rst
sjanssen2: wow. Jira is constantly consuming 100% of one of my CPU cores :-1: 
Sure that there is no further configuration of Jira necessary?

antgonza: AFAIK, nothing else. Suggest contacting the JIRA devs ...

README.rst
```diff
@@ -9,7 +9,7 @@ Installation instructions
 
 Create a conda environment for labadmin::
 
-   conda create -n labadmin psycopg2 python=2 click requests
+   conda create -n labadmin psycopg2 python=2 click requests bioconda java-jdk
```
antgonza: @sjanssen2, do you need bioconda here?

sjanssen2: I don't if you find a java-jdk osx / lx64 package in the default channels

antgonza: could you find out? like I said, I have never used conda ...

sjanssen2: no. I don't have the time for this right now. @EmbrietteH needs me to work on the metadata pulldown. My google hit's indicated that the best channel is bioconda. But there might be more out there.
I thought using conda is kind of the lab standard ?!

josenavas: @sjanssen2 apparently you didn't had a problem installing this with the current instructions, right? If so, let's move forward with this and not block this PR. On the other hand, if you have installed this using conda, simply post your commands and @antgonza will update the PR with your commands.

sjanssen2: nope. I couldn't follow instructions, because my java on Linux Mint is < 1.8, thus I needed to use a conda env, but I might be a special case since I am not using a Mac. Thus, for the default (Mac) user the instructions might work. I simply cannot test them :-/

antgonza: k, decided to split the conda command in 2, which is the easiest/fastest.

josenavas: Thanks @antgonza

antgonza:
@sjanssen2 and @josenavas, what do you think?

josenavas:
@sjanssen2 please merge if you agree with this last version.

Thanks!

PR #192, opened 2017-04-06
Fixing all methods in data access

josenavas:
I noticed that my PR was growing too much. I end up needing to fix all the methods in data access (rather than a few as I was expecting) so I've decided to just first do a PR with the modifications on the data access file. This is a bit bigger than usual but is the smallest that I can make it w/o duplicating too much effort.

@qiyunzhu as a big change from the previous structure that you had in place is that now the samples must belong to one study, so no need to have independent functions to interact with the samples individually.

### DIFF
--- a/knimin/lib/data_access.py
+++ b/knimin/lib/data_access.py
@@ -1,6 +1,7 @@
 from __future__ import unicode_literals
 from contextlib import contextmanager
 from collections import defaultdict, namedtuple
+from itertools import chain
 from os import walk
 from os.path import join, splitext, isdir, abspath
 from copy import copy
@@ -2618,9 +2619,6 @@ class KniminAccess(object):
     def delete_study(self, study_id):
         """Deletes an existing study
 
-        Samples exclusively associated with the study will be deleted. Samples
-        associated with it but also with other studies will be disassociated.
-
         Parameters
         ----------
         study_id : int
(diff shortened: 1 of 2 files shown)


josenavas:
Failures are expected until this is merged https://github.com/biocore/american-gut-web/pull/653

knimin/lib/data_access.py
```diff
@@ -2630,283 +2628,120 @@ def delete_study(self, study_id):
...
-                if res:
-                    raise ValueError('Sample ID(s) %s already exist.' %
-                                     ', '.join(sorted(res)))
-
-    def _validate_samples(self, samples):
-        """Validate sample properties
-
-        Confirms that the barcodes and study IDs to be assigned to samples are
-        valid.
+    def set_study_samples(self, study_id, samples):
```
qiyunzhu: I am bit unclear with the logics behind this function. What are all possible mechanism(s) of adding samples to labadmin? Here I see one mechanism: the user selects samples associated with a Qiita study in the plate mapper interface. Therefore, "create a sample" === "plate a sample".

Then, in what situations will samples be dissociated from a plate & removed? In the docstring of this function I learned that a plated sample cannot be removed.

It will be great if you can add some explanations.

knimin/lib/data_access.py
```diff
@@ -2630,283 +2628,120 @@ def delete_study(self, study_id):
...
-            If one or more samples are associated with one or more sample
-            plates, in which case, the sample plates have to be deleted
-            prior to the deletion of the samples
+            If the study ID does not exist
         """
         with TRN:
-            self._samples_exist(ids)
-            sql = """SELECT sample_plate_id, sample_id
+            self._study_exists(study_id)
+            sql = """SELECT sample_plate_id,
```
qiyunzhu: If I understand correctly, this SQL query will return plate IDs (`res[0]`) and sample IDs (`res[1]`) that are 1) on those plates and 2) belong to the given study ID. Meanwhile, Samples that are 1) on those plates and 2) belong to other studies will not be returned. Is that correct? If the purpose of this function is to render existing plate(s), there may be some empty wells.

knimin/lib/data_access.py
```diff
@@ -3165,96 +3031,130 @@ def read_sample_plate(self, plate_id):
...
         """
         with TRN:
             sql = """DELETE FROM pm.sample_plate_layout
                      WHERE sample_plate_id = %s"""
-            TRN.add(sql, [id])
+            TRN.add(sql, [sample_plate_id])
             TRN.execute()
 
-    def write_sample_plate_layout(self, id, layout):
+    def write_sample_plate_layout(self, sample_plate_id, layout):
```
qiyunzhu: How are empty wells dealt with in this function?

knimin/lib/data_access.py
```diff
@@ -3165,96 +3031,130 @@ def read_sample_plate(self, plate_id):
...
+            for row_idx, row in enumerate(layout):
+                for col_idx, values in enumerate(row):
+                    sql_args.append([sample_plate_id, values['sample_id'],
+                                     row_idx, col_idx, values['name'],
+                                     values['notes']])
+            TRN.add(sql, sql_args, many=True)
+            TRN.execute()
 
-    def read_sample_plate_layout(self, id):
+    def read_sample_plate_layout(self, sample_plate_id):
```
qiyunzhu: I have no doubt that it will work, but I just feel that it's overly complicated.

knimin/lib/data_access.py
```diff
@@ -3353,48 +3275,41 @@ def get_sample_plate_list(self):
...
-                                  COUNT(sample_id) AS sample_count
-                           FROM pm.study
-                           JOIN pm.study_sample USING (study_id)
-                           JOIN pm.sample_plate_layout USING (sample_id)
-                           JOIN pm.sample_plate USING (sample_plate_id)
-                           GROUP BY study_id, sample_plate_id
-                           ORDER BY COUNT(study_id) DESC) AS x
-                     USING (sample_plate_id)
-                     ORDER BY sample_plate_id"""
+            sql = """SELECT sp.sample_plate_id as id,
```
qiyunzhu: I am very impressed that you can make such a comprehensive SQL query!

knimin/lib/tests/test_data_access.py
```diff
@@ -500,200 +501,180 @@ def test_delete_study(self):
...
+        row = []
+        for i in range(pt['rows']):
+            for j in range(pt['cols']):
+                row.append({'sample_id': None, 'name': None, 'notes': None})
+            layout.append(row)
+            row = []
+        layout[0][0]['sample_id'] = '9999.sample1'
+        layout[0][1]['sample_id'] = '9999.sample2'
+        db.write_sample_plate_layout(plate_id, layout)
+        samples.remove('9999.sample2')
```
qiyunzhu: This line doesn't seem to do anything for the test?

knimin/lib/data_access.py
josenavas: I've removed the function "create a sample". The only way of adding a sample to the system at this moment is using this function. This function is not directly used by any user, but triggered by the system when sync-ing the studies in the DB with Qiita. Note that the only "recognized" samples are the samples that are present in Qiita. So, although you can remove a plate, the sample is not necessarily removed from the system. The sample is removed from the system if it is removed from Qiita. Does that make sense?

knimin/lib/data_access.py
```diff
@@ -2630,283 +2628,120 @@ def delete_study(self, study_id):
...
         """
         with TRN:
             self._study_exists(study_id)
-            samples = self.get_samples_by_study(study_id)
+            # Check if there is any sample that has been already plated
+            plated = set(chain.from_iterable(
+                self.get_study_plated_samples(study_id).values()))
+            if plated:
+                raise ValueError(
+                    "Can't remove study %s, samples have been plated"
```
antgonza: not sure but perhaps add ... "try removing the plates first"?

knimin/lib/data_access.py
```diff
@@ -2630,283 +2628,120 @@ def delete_study(self, study_id):
...
-                if any(x in sample for x in ['is_blank', 'barcode', 'notes']):
-                    TRN.add(sql1, [sample.get('is_blank', False),
-                                   sample.get('barcode'), sample.get('notes'),
-                                   sample['id']])
-                if 'study_ids' in sample:
-                    TRN.add(sql2, [sample['id']])
-                    TRN.add(sql3, [sample['id'], sample['study_ids']])
-            TRN.execute()
+            self._study_exists(study_id)
+            # Check if there are any samples that where already plated that
```
antgonza: where -> were

knimin/lib/data_access.py
```diff
@@ -2630,283 +2628,120 @@ def delete_study(self, study_id):
...
+            # are being removed
+            plated = set(chain.from_iterable(
+                self.get_study_plated_samples(study_id).values()))
+            samples = set(samples)
+            removed = plated - samples
+            if removed:
+                raise ValueError(
+                    'Plated samples have been removed from the study: %s'
```
antgonza: I'm not sure this message makes sense, can you clarify?

knimin/lib/data_access.py
antgonza: @qiyunzhu, I guess that's exactly the point of this function: "get_study_plated_samples" but let's see what @josenavas has to say ...

knimin/lib/data_access.py
```diff
@@ -3165,96 +3031,130 @@ def read_sample_plate(self, plate_id):
...
-            sql = """INSERT INTO pm.sample_plate_layout (sample_plate_id,
-                        sample_id, col, row, name, notes)
+            self._sample_plate_exists(sample_plate_id)
+
+            # Check if the given layout forms a valid layout
+            l_rows = len(layout)
+            l_cols = len(layout[0])
+            # Check that all rows have the same number of columns
+            # for r in layout
+            if any([len(r) != l_cols for r in layout]):
```
antgonza: @qiyunzhu from this line I think all wells have to be there to proceed ...

knimin/lib/data_access.py
antgonza: Could you add more background? What's complicated? Also perhaps adding a better solution will be helpful

knimin/lib/data_access.py
```diff
@@ -3315,6 +3215,28 @@ def get_plate_types(self):
             TRN.add(sql)
             return [dict(x) for x in TRN.execute_fetchindex()]
 
+    def read_plate_type(self, plate_type_id):
```
antgonza: Should this return the string name of the read_plate?

knimin/lib/data_access.py
```diff
@@ -3315,6 +3215,28 @@ def get_plate_types(self):
...
+        plate_type_id: int
+            The id of the plate type
+
+        Returns
+        -------
+        DictCursor
+            The information of the plate type
+        """
+        with TRN:
+            sql = "SELECT * FROM pm.plate_type WHERE plate_type_id = %s"
```
antgonza: Can you be explicit about what you are requesting?

knimin/lib/data_access.py
```diff
@@ -3165,96 +3031,130 @@ def read_sample_plate(self, plate_id):
...
+            A 2-D matrix storing the per well information in a dict with the
+            format: {'sample_id': str, 'name': str, 'notes': str}
         """
         with TRN:
-            self._sample_plate_exists(id)
-            if not self._sample_plate_layout_exists(id):
-                return []
-            sql = """SELECT sample_id, col, row, name, notes
+            self._sample_plate_exists(sample_plate_id)
+            sql = """SELECT *
```
antgonza: Can you be explicit about what you are expecting?

knimin/lib/data_access.py
antgonza: ... me too ... 😉

knimin/lib/tests/test_data_access.py
```diff
@@ -500,200 +501,180 @@ def test_delete_study(self):
...
+
+        # Try to get plated samples from a study that does not exist
+        with self.assertRaises(ValueError) as ctx:
+            db.get_study_plated_samples(9998)
+        self.assertEqual(ctx.exception.message,
+                         'Study ID 9998 does not exist.')
+
+    def test_get_study_samples(self):
+        # Create a study
+        db.create_study(9999, title='LabAdmin test project', alias='LTP',
```
antgonza: just remember that the jira_id is not the same than the jira_key ... but just double checked the documentation and you can instantiate a jira project either by key or id ...

knimin/lib/data_access.py

knimin/lib/data_access.py
josenavas: done

knimin/lib/data_access.py
josenavas: Changed to:
"Can't remove samples from study %s. Offending samples: %s"

knimin/lib/data_access.py
josenavas: That's correct. However, this function will not be used to render existing plates, but for checking which samples from the study have been already plated.

The function used to render the existing plates is `read_sample_plate_layout`

knimin/lib/data_access.py
josenavas: In the empty well, the dictionary will have have None values in every key.

knimin/lib/data_access.py
josenavas: Correct, wells need to be defined, but just need to pass None values to identify an empty well

knimin/lib/data_access.py
josenavas: Note that the representation changed. Instead of being a list of dicts with 'col', 'row', 'sample', etc, the current representation is just a dense 2D matrix storing this values. I think this representation makes more sense given than in general the plates tend to be dense rather than sparse, so a matrix representation is more efficient.

knimin/lib/data_access.py

knimin/lib/data_access.py
josenavas: I don't think I understand this comment.

knimin/lib/data_access.py

knimin/lib/data_access.py
josenavas: To be fair - it probably took me 1 hour to get it right hahahaha

knimin/lib/tests/test_data_access.py
josenavas: Good catch! Removed

knimin/lib/tests/test_data_access.py
josenavas: oh cool - `jira_id` is stored as a varchar, so we can use either one in there 😄

knimin/lib/data_access.py
antgonza: The question is: do we have an actual string for plate_type_id or no? If yes, should we return that string vs. the id?

knimin/lib/data_access.py
```diff
@@ -3165,96 +3031,128 @@ def read_sample_plate(self, plate_id):
...
             TRN.execute()
 
-    def write_sample_plate_layout(self, id, layout):
+    def write_sample_plate_layout(self, sample_plate_id, layout):
         """Writes the layout of a sample plate
 
         Parameters
         ----------
-        id : int
+        sample_plate_id : int
```
tanaes: Because plates correspond to actual physical objects, does it make sense to require the plate_type attribute be specified explicitly here?  Or has that already been specified earlier on in the plate definition, and that's what's your querying on line 3080?

knimin/lib/data_access.py
antgonza: After offline discussion, I'm OK with this.

knimin/lib/data_access.py
```diff
@@ -3165,96 +3031,128 @@ def read_sample_plate(self, plate_id):
...
-            return [dict(x) for x in TRN.execute_fetchindex()]
+                     ORDER BY row, col"""
+            TRN.add(sql, [sample_plate_id])
+            res = TRN.execute_fetchindex()
+            layout = []
+            row = []
+            # To construct the output layout we just need to iterate over
+            # the results since they are correctly sorted
+            for well in res:
+                # Transform the DictCursos to an actual dictionary
```
tanaes: Is this DictCursors? 

I didn't know about DictCursors so I was wondering what a DictCurso was. :D #spanishpython

josenavas:
Feels good to see an increase of coverage of 2.6%!! @qiyunzhu can you take another look and merge if you agree with the changes? thanks!

knimin/lib/tests/test_data_access.py
```diff
@@ -747,49 +729,90 @@ def test_create_sample_plate(self):
...
+        row = []
+        for i in range(pt['rows']):
+            for j in range(pt['cols']):
+                row.append({'sample_id': None, 'name': None, 'notes': None})
+            layout.append(row)
+            row = []
+        layout[0][0]['sample_id'] = samples[0]
+        layout[0][1]['sample_id'] = samples[1]
+        layout[0][2]['sample_id'] = samples[2]
+        db.write_sample_plate_layout(plate_id, layout)
```
tanaes: Are these lines (795-809) actually part of this test? I don't see an assert anywhere...

knimin/lib/data_access.py
```diff
@@ -2630,283 +2628,120 @@ def delete_study(self, study_id):
...
-                    TRN.add(sql2, [sample['id']])
-                    TRN.add(sql3, [sample['id'], sample['study_ids']])
-            TRN.execute()
+            self._study_exists(study_id)
+            # Check if there are any samples that were already plated that
+            # are being removed
+            plated = set(chain.from_iterable(
+                self.get_study_plated_samples(study_id).values()))
+            samples = set(samples)
+            removed = plated - samples
```
josenavas: In general I would agree with you. However, in this case the user is not going to actively add/remove samples. This function is being used to sync the samples in LabAdmin and in Qiita, so the only thing that we receive is a set of samples, w/o distinguishing which have been added and which have been removed. Thus I find that this is the best way of doing it.

knimin/lib/data_access.py
tanaes: I was pondering this a bit and I think I agree. The downside of the dense matrix representation here is that it's more likely to have errors due to the implicit specification of wells -- so if you accidentally ordered something incorrectly you might end up with A samples ending up in H wells. So it's probably worth keeping that in mind when writing verifications. But the advantage is that you can't accidentally wind up partially filling a larger plate format.

knimin/lib/data_access.py
josenavas: The plate_type have been specified on plate definition so here I just wanted to provide the option to modify the plate_type just before the user starts plating, but in general I do not expect this changing.

knimin/lib/data_access.py
josenavas: Yeah is [DictCursors](http://initd.org/psycopg/docs/extras.html#dictionary-like-cursor)

Typing too fast end up on savng lettrs 😄

knimin/lib/tests/test_data_access.py
josenavas: Yes, what these lines are making is forcing the error raised on line 550. These lines are plating some samples for the study 9999, so when the user tries to remove the study it fails because there are samples already plated.

knimin/lib/data_access.py
josenavas: Yeah, I do agree that you need to be careful when providing the order of the wells. The advantage, however, is that the 2D representation is kept consistent from the interface to here. In fact, since the user is already seeing a 2D matrix when plating I found it was harder to mess up if I keep the same representation all the time.

tanaes: Yep totally agree! 👍

knimin/lib/data_access.py
tanaes: Ok that makes sense to me.

knimin/lib/tests/test_data_access.py
tanaes: Ahh I see. Thanks!

knimin/lib/data_access.py
qiyunzhu: @josenavas Got it!

knimin/lib/data_access.py
qiyunzhu: @josenavas Sounds good!

PR #193, opened 2017-04-06
age is now based on collection timestamp not on now

sjanssen2:
Age of participant was calculated as the difference between their birthday and now(), i.e. the time of the pulldown, which is wrong. What we want to do is base it off the collection timestamp.
This tiny PR will fix this problem.
Thanks @jwdebelius for spotting this!


sjanssen2:
ready for review @josenavas @antgonza @wasade

jwdebelius:
Code looks good to me; I dont know the binaries well enough to evaluate whether the update makes sense there.

PR #194, opened 2017-04-06
WIP: vioscreen API

sjanssen2:
DON'T MERGE FOR NOW.

added vioscreen API login info to the configuration. Default are for their demo server. Actual credentials can be obtained from Embriette.

### DIFF
--- a/knimin/config.txt.example
+++ b/knimin/config.txt.example
@@ -29,3 +29,9 @@ HOST = http://127.0.0.1:2990/jira
 USERNAME = admin
 PASSWORD = admin
 PASSKEY =
+
+[vioscreen]
+HOST = https://demo.api.viocare.com
+USERNAME = KLDEMO
+PASSWORD = KLdemo17
+REGCODE = KLUCB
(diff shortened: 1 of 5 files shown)


sjanssen2:
@EmbrietteH @jwdebelius
My basic module is now able to retrieve data from vioscreen through their API.
For each query, I first need to obtain the "sessions" (the times a user filled out the questionnaire) for a given user (user corresponds to our survey_id). This typically returns 1 session. For this session, I collect all stored information via 7 separate API calls. Thus, it takes approx 8 seconds to get a full dataset for one survey_id.

Do we expect that a user goes back to his survey and changes information in it? If not, we might want to use a caching mechanism, such that we only gather information for a survey_id once.
Alternatively - but I guess Vioscreen guys won't like it - we might ask them to provide a specialized API function to batch-download all sessions at once. 

Each dataset is ~1.7MB of information. I don't know if it is really necessary to store all of them. Can we sit together, go through the list of fields and decide which are not needed?

jwdebelius:
If a survey is complete, the results should not change.

If a survey is incomplete, than it could be completed. I don't think a lot of people go back and compete their surveys, but its possible. When I use the GUI, the incomplete surveys don't show up for me to download; I don't know how it works for the API?

So, I think that once the vioscreen data is downloaded, you should only need to download new surveys, rather than pulling down the data repeatedly.

sjanssen2:
Thanks Justine. I did a quick check with the productive dataset and found that:
2 users have vioscreen status "New",
4 are "Started",
the mentioned ~1700 are status "Finished" 
and the remaining have no Vioscreen at all

wasade:
Superseded by #247

PR #198, opened 2017-04-07
Extract plate page

josenavas:
This adds the page for choosing the parameters during plate extraction.

The error at end is related with the next page, which I'll be fixing in another PR.

extractplate

### DIFF
--- a/knimin/lib/data_access.py
+++ b/knimin/lib/data_access.py
@@ -3197,6 +3197,59 @@ class KniminAccess(object):
                                prop + '_id'))
             return [dict(x) for x in TRN.execute_fetchindex()]
 
+    def get_or_create_property_option_id(self, prop, option_name):
+        """Retrieves the id of the given property
+
+        If the property with the given name doesn't exist, it will create it
+        and return the id of the newly created value
+
+        Parameters
+        ----------
+        prop : str
+            Property name, i.e., name of a table under schema "pm"
+        option_name : str
+            The new option
+        """
+        prop_id_col = prop + '_id'
+        with TRN:
+            # Insert if it doesn't exist
+            sql = """INSERT INTO pm.{} (name)
+                        SELECT %s
+                        WHERE NOT EXISTS(SELECT 1 FROM pm.{} WHERE name = %s)
+                        RETURNING {}""".format(prop, prop, prop_id_col)
+            TRN.add(sql, [option_name, option_name])
+            res = TRN.execute_fetchindex()
+            if res:
+                # This means that the previous sql call inserted a new value
+                # so we just need to return the value
+                # Magic numbers [0][0] - There is only one row and with
+                # a single value
+                return res[0][0]
+
+            # If the code reaches this point it means that the previous sql
+            # call did not insert anything, i.e. the option already existed
+            # return the id of that option
+            sql = "SELECT {} FROM pm.{} WHERE name = %s".format(
+                prop_id_col, prop)
+            TRN.add(sql, [option_name])
+            return TRN.execute_fetchlast()
+
+    def delete_property_option(self, prop, prop_id):
+        """Deletes the given property option
+
+        Parameters
+        ----------
+        prop : str
+            Property name, i.e., name of a table under schema "pm"
+        prop_id : int
+            The id of the property to remove
+        """
+        prop_id_col = prop + '_id'
+        with TRN:
+            sql = "DELETE FROM pm.{} WHERE {} = %s".format(prop, prop_id_col)
+            TRN.add(sql, [prop_id])
+            TRN.execute()
+
     def get_plate_types(self):
         """Gets all available plate types
 
--- a/knimin/lib/data_access.py
+++ b/knimin/lib/data_access.py
@@ -3312,6 +3312,33 @@ class KniminAccess(object):
                 plates.append(row)
             return plates
 
+    def extract_sample_plates(self, sample_plate_ids, email, robot_id, kit_id,
+                              tool_id, notes=None):
+        """Stores the extraction information for the given sample_plates
+
+        Parameters
+        ----------
+        sample_plate_ids : list of int
+            The sample plates being extracted
+        email : str
+            The email of the user preparing the extraction
+        robot_id : str
+            The id of the robot used for extraction
+        kit_id : str
+            The id of the kit lot used for extraction
+        tool_id : str
+            The id of the tool used for extraction
+        notes : list of str, optional
+            Notes added to the extracted plates
+        """
+        with TRN:
+            sql = """INSERT INTO pm.dna_plate (
+                        email, created_on, sample_plate_id,
+                        extraction_robot_id, extraction_kit_lot_id,
+                        extraction_tool_id, notes)
+                     VALUES (%s, now(), %s, (%s), (%s), (%s), %s)"""
+            print sql
+
     def _clear_table(self, table, schema):
         """Test helper to wipe out a database table"""
         self._con.execute('DELETE FROM %s.%s' % (schema, table))
--- a/knimin/lib/data_access.py
+++ b/knimin/lib/data_access.py
@@ -3365,8 +3365,8 @@ class KniminAccess(object):
                 plates.append(row)
             return plates
 
-    def extract_sample_plates(self, sample_plate_ids, email, robot_id, kit_id,
-                              tool_id, notes=None):
+    def extract_sample_plates(self, sample_plate_ids, email, robot, kit_lot,
+                              tool, notes=None):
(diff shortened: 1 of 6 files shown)


josenavas:
@ElDeveloper @qiyunzhu @sjanssen2 are you available for a review?

knimin/templates/pm_extract_plate.html
```diff
@@ -0,0 +1,186 @@
...
+      var plateId = $("#plate-select").val();
+      var plateName = $("#plate-select option[value=" + plateId + "]").attr('pm-name');
+
+      // Add the new plate to the list
+      var $li = $('<li>').addClass('list-group-item justify-content-between').attr('id', 'li-plate-' + plateId).attr('pm-id', plateId).html('(ID: ' + plateId + ') ' + plateName).appendTo($("#lu-plate-list"));
+      var $btn = $('<button>').addClass('btn-danger glyphicon glyphicon-remove pull-right').html(" ").appendTo($li);
+      $btn.attr('pm-name', plateName).attr('pm-id', plateId);
+      $btn.click(function (e) {
+        // Add this plate back to the select
+        $("#plate-select").append($('<option>', {value: $(this).attr('pm-id'),
```
qiyunzhu: There are `append` in some cases and `appendTo` in others. I am not familiar with either. Just wonder if they behave differently somehow?

knimin/lib/data_access.py
```diff
@@ -3312,6 +3365,118 @@ def get_sample_plate_list(self):
...
+
+        Parameters
+        ----------
+        dna_plate_id : int
+            The id of the DNA plate
+
+        Raises
+        ------
+        ValueError
+            If the dna plate does not exist
```
qiyunzhu: `dna` => `DNA`

knimin/lib/data_access.py
```diff
@@ -3312,6 +3365,118 @@ def get_sample_plate_list(self):
...
+            The name of the robot used for extraction
+        kit_lot : str
+            The kit lot used for extraction
+        tool : str
+            The name of the tool used for extraction
+        notes : str, optional
+            Notes added to the extracted plates
+        """
+        if not sample_plate_ids:
+            raise ValueError("Provide at least one sample plate to extract")
```
qiyunzhu: This error is not mentioned in the docstring. Also, in what case a `sample_plate_ids` can be `[]`?

knimin/lib/data_access.py
```diff
@@ -3197,6 +3197,59 @@ def get_property_options(self, prop):
                                prop + '_id'))
             return [dict(x) for x in TRN.execute_fetchindex()]
 
+    def get_or_create_property_option_id(self, prop, option_name):
```
qiyunzhu: This function does two things, which can be separated into two functions. Will it be possible that the user wants to create a new property (extraction kit lot?) not with a plate, but somehow _de novo_?

knimin/templates/pm_extract_plate.html
```diff
@@ -0,0 +1,186 @@
...
+    $("#extraction-kit-input").removeClass('pm-warning');
+    if(val.length > 0 && $.inArray(val, kits) === -1) {
+      $("#extraction-kit-input").addClass('pm-warning');
+    }
+  }
+
+  $(document).ready(function() {
+    var plates = {% raw [ [int(p[0]), p[1]] for p in plates ] %};
+
+    // I need this modal to be added to the body, otherwise it doesn't behave
```
qiyunzhu: Maybe in this block of code you mentioned some DOM that does not exist in HTML but is created dynamically via JavaScript?

knimin/lib/data_access.py
josenavas: I think this function can be used if the user wants to create a property _de novo_ , since it will create it if it doesn't exist and just return the id, I don't think there is any need to have a plate on this function

knimin/lib/data_access.py
josenavas: Good catch! Added to the docstring

In theory, if this is used through the interface we should never hit this error. However, if there is a bug somewhere or this function ends up being used somewhere else, the error generated by the DB when executing this function is ugly, so at least here we have a useful error message.

knimin/lib/data_access.py

knimin/templates/pm_extract_plate.html
```diff
@@ -0,0 +1,186 @@
...
+
+  /**
+   *
+   * Formats the background color of the kit lot select
+   *
+   **/
+  function formatKitLot(kits) {
+    var val = $("#extraction-kit-input").val().trim();
+    $("#extraction-kit-input").removeClass('pm-warning');
+    if(val.length > 0 && $.inArray(val, kits) === -1) {
```
josenavas: Added!

knimin/templates/pm_extract_plate.html
josenavas: The "modal" is the div created in the following line (it's a div, but the class is "modal").

Modal it's a bootstrap class that allows to show what looks like a pop-up window but prettier (e.g. in the gif is used to show the plate selection). However, in order for it to work correctly, it needs to be attached to the body directly. The block content is in line 148, which is used by the tornado template system. That block is defined inside a div in another page (auth_sitebase.html).

For reference, this is the [issue](http://stackoverflow.com/q/10636667) that I'm seeing. Although I could potentially modify auth_sitebase.html, it can potentially impact so many pages that I don't feel comfortable making that modification ATM.

knimin/templates/pm_extract_plate.html
josenavas: Yes, they behave differently.
`A.append(B)` Appends the element B to A
`A.appendTo(B)` Appends the element A to B
I use one or the other based on how I can save using variables and based on what I'm creating at that moment.

knimin/templates/pm_extract_plate.html
qiyunzhu: Thanks for explaining the mechanism! I didn't think that it is so complicated.

knimin/templates/pm_extract_plate.html
qiyunzhu: Sounds good. Thanks!

knimin/lib/data_access.py
qiyunzhu: Makes sense!

knimin/lib/data_access.py
qiyunzhu: Makes sense!

qiyunzhu:
All look good now! Thanks for explaining to me the techs you used!

josenavas:
Not a problem! I'fe you don't know about [bootstrap](http://getbootstrap.com/) and you like to do web-based development, I highly recommend to take a look at it! It simplifies a lot of tasks and makes super-easy to make really cool-looking pages!

qiyunzhu:
@josenavas Thanks! I know a little bit of Bootstrap. Will explore!

knimin/templates/pm_extract_plate.html
```diff
@@ -0,0 +1,186 @@
...
+  function formatKitLot(kits) {
+    var val = $("#extraction-kit-input").val().trim();
+    $("#extraction-kit-input").removeClass('pm-warning');
+    if (val.length > 0 && $.inArray(val, kits) === -1) {
+      $("#extraction-kit-input").addClass('pm-warning');
+    }
+  }
+
+  $(document).ready(function() {
+    var plates = {% raw [ [int(p[0]), p[1]] for p in plates ] %};
```
ElDeveloper: Instead of formatting the list, perhaps it would be more convenient to make sure the data is in a JSON-compatible format and then just call json.dumps.

knimin/templates/pm_extract_plate.html
```diff
@@ -0,0 +1,186 @@
...
+                                                 "pm-name": $(this).attr('pm-name')}));
+        $li.remove();
+      });
+
+      // Set the selected value to -1
+      $("#plate-select").val("-1");
+      // Delete the selected plate from the option list
+      $("#plate-select option[value=" + plateId + "]").remove();
+      // Hide the modal
+      $('#addPlateModal').modal('hide');
```
ElDeveloper: Minor but addPlateModal is the only camelCased identifier, maybe for consistency worth changing to add-plate-modal

knimin/templates/pm_extract_plate.html
josenavas: Good call! The list is already JSON-compatible so I've changed it to use dumps

knimin/templates/pm_extract_plate.html
josenavas: Done

PR #199, opened 2017-04-09
Fix list and platemapper.js

josenavas:
~~Depends on #198 - review/merge that one first~~

This PR:
- fixes the list page (small changes)
- modifies platemapper.js so it can be prepopulated with the info from the DB
- modifies platemapper.js to detect technical replicates in the same plate
- disables the extract button if there are errors in the plate

Gif:
platelist

### DIFF
--- a/knimin/handlers/pm_plate_map.py
+++ b/knimin/handlers/pm_plate_map.py
@@ -125,26 +125,22 @@ class PMExtractPlateHandler(BaseHandler):
         plate_id = self.get_argument('plate_id')
         plate_name = db.read_sample_plate(plate_id)['name']
         plates = [[p['id'], p['name']] for p in db.get_sample_plate_list()]
+        robots = db.get_property_options('extraction_robot')
+        tools = db.get_property_options('extraction_tool')
+        kits = db.get_property_options('extraction_kit_lot')
 
         self.render("pm_extract_plate.html", currentuser=self.get_current_user,
-                    plate_id=plate_id, plate_name=plate_name, plates=plates)
+                    plate_id=plate_id, plate_name=plate_name, plates=plates,
+                    robots=robots, tools=tools, kits=kits)
 
-    # @authenticated
-    # def post(self):
-    #     plate_id = self.get_argument('plate_id')
-    #     action = self.get_argument('action')
-    #     layout = json_decode(self.get_argument('layout'))
-    #
-    #     if action not in ('save', 'extract'):
-    #         raise HTTPError(400, 'Action should be save or extract')
-    #     else:
-    #         # In any of the two cases we need to save the plate layout
-    #         db.write_sample_plate_layout(plate_id, layout)
-    #
-    #         if action == 'save':
-    #             # At this point we are done! Simply return a 200
-    #             self.set_status(200)
-    #             self.finish()
-    #         elif action == 'extract':
-    #             # The plate is promoted for extraction
-    #             self.redirect("/pm_extract_plate?plate_id=%s" % plate_id)
+    @authenticated
+    def post(self):
+        plates = json_decode(self.get_argument('plates'))
+        robot = self.get_argument('robot')
+        tool = self.get_argument('tool')
+        kit = self.get_argument('kit')
+        user = self.current_user
+
+        db.extract_sample_plates(plates, user, robot, kit, tool)
+
+        self.redirect("/pm_plate_list/")
--- a/knimin/handlers/pm_plate_map.py
+++ b/knimin/handlers/pm_plate_map.py
@@ -107,12 +107,17 @@ class PMSamplePlateHandler(BaseHandler):
             study['samples'] = {}
             study['samples']['all'] = db.get_study_samples(s_id)
             study['samples']['plated'] = db.get_study_plated_samples(s_id)
+            # Remove the entry for the current plate if it exists
+            # Note: casting to long because the keys in the dictionary are
+            # longs, so we need the cast to safely remove the entry
+            study['samples']['plated'].pop(long(plate_id), None)
             studies.append(study)
 
         plate_info['studies'] = studies
 
         plate_info['plate_id'] = plate_id
         plate_info['created_on'] = plate_info['created_on'].isoformat(sep=' ')
+        plate_info['layout'] = db.read_sample_plate_layout(plate_id)
 
         self.write(plate_info)
         self.finish()
--- a/knimin/handlers/pm_plate_map.py
+++ b/knimin/handlers/pm_plate_map.py
@@ -123,10 +123,11 @@ class PMExtractPlateHandler(BaseHandler):
     @authenticated
     def get(self):
         plate_id = self.get_argument('plate_id')
+        plate_name = db.read_sample_plate(plate_id)['name']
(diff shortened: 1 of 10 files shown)


josenavas:
Ready for review!

knimin/templates/pm_plate_list.html
```diff
@@ -133,7 +129,7 @@ <h3 id="page-title">{{target[0].upper() + target[1:]}} Plate List</h3>
           <i class="fa fa-battery-empty"></i>
           {% end %}
         </span></td>
-      <td title="{{plate['study'][3]}}">{{plate['study'][1]}}</td>
+      <td title="{{plate['studies'][0]}}">{{plate['studies'][0]}}</td>
```
antgonza: Can you add a comment about the magic numbers? -- From the current context is not clear why [3] -> [0] and [1] to [0]

knimin/static/js/platemapper.js
```diff
@@ -269,36 +303,137 @@ PlateMap.prototype.keypressWell = function (current, e) {
...
+  } else {
+    // Using a for loop to avoid to special case - If the wells list is
+    // empty, this for loop will not execute anything. If there is one element
+    // it will correctly remove the mark for techincal replicates
+    for (var well of sampleInfo.wells) {
+      this.wellInformation[well[0]][well[1]].inputTag.removeClass('pm-technical-replicate');
+    }
+    // Remove from the techincal replicates list
+    this.technicalReplicates = $.grep(this.technicalReplicates, function(value) {
+      return value != sampleInfo;
```
qiyunzhu: The return value doesn't seem to be used in elsewhere.

knimin/static/js/platemapper.js
```diff
@@ -269,36 +303,137 @@ PlateMap.prototype.keypressWell = function (current, e) {
...
+ **/
+PlateMap.prototype._updateTechnicalReplicate = function (sampleInfo) {
+  if (sampleInfo.wells.length > 1) {
+    // Update the list of technical replicates
+    if ($.inArray(sampleInfo, this.technicalReplicates) == -1) {
+      this.technicalReplicates.push(sampleInfo);
+    }
+    // Mark all wells as technical
+    for (var well of sampleInfo.wells) {
+      this.wellInformation[well[0]][well[1]].inputTag.addClass('pm-technical-replicate');
```
wasade: Do the index positions into well correspond to row/col?

knimin/handlers/pm_plate_map.py
```diff
@@ -107,12 +107,17 @@ def get(self):
             study['samples'] = {}
             study['samples']['all'] = db.get_study_samples(s_id)
             study['samples']['plated'] = db.get_study_plated_samples(s_id)
+            # Remove the entry for the current plate if it exists
+            # Note: casting to long because the keys in the dictionary are
+            # longs, so we need the cast to safely remove the entry
+            study['samples']['plated'].pop(long(plate_id), None)
```
qiyunzhu: I didn't know that JavaScript can `pop` a specific element from array :)

knimin/static/js/platemapper.js
```diff
@@ -269,36 +303,137 @@ PlateMap.prototype.keypressWell = function (current, e) {
...
+ *
+ **/
+PlateMap.prototype.updateErrorList = function () {
+  var well, message, $li;
+  // Empty the current list
+  $("#pm-error-list").html("");
+
+  // First look for any warning/error in a per well basis
+  for (var i = 0; i < this.wellInformation.length; i++) {
+    for (var j = 0; j < this.wellInformation[i].length; j++) {
```
wasade: Is indexing necessary here? Isn't the length of the plate fixed?

knimin/static/js/platemapper.js
```diff
@@ -113,11 +116,42 @@ PlateMap.prototype.initialize = function (data) {
...
+    }
 };
 
 /**
  *
+ * Populates the GUI with the plate information
+ *
+ * @param {Array} layout The plate information
+ **/
+PlateMap.prototype._populatePlate = function (layout) {
```
qiyunzhu: This will also restore the warning / error informations, will it?

knimin/static/js/platemapper.js
```diff
@@ -269,36 +303,137 @@ PlateMap.prototype.keypressWell = function (current, e) {
...
+  prevSampleInfo = this.samples[$(current).data('pm-sample')];
+  // Store the well Id as a list with 2 elements: the row # and the col #
+  // This will simplify updating the interface
+  wellId = [parseInt($(current).attr('pm-well-row')),
+            parseInt($(current).attr('pm-well-column'))];
+  if (prevSampleInfo !== undefined) {
+    // We had another sample in this well before, update the wells list
+    // of that sample
+    prevSampleInfo.wells = $.grep(prevSampleInfo.wells, function(value) {
+      return (value[0] != wellId[0]) || (value[1] != wellId[1]);
```
wasade: Should these be !==?

knimin/static/js/platemapper.js
```diff
@@ -269,36 +303,137 @@ PlateMap.prototype.keypressWell = function (current, e) {
...
+        $li = $('<li>').addClass('list-group-item').addClass('list-group-item-warning').html(message).appendTo('#pm-error-list');
+        // Add a link for each plate that the sample is already plated
+        for (var plate of this.samples[sample].plates) {
+          $('<a>').attr('href', '/pm_plate_map?plate_id=' + plate).attr('target', '_blank').text(plate + ' ').appendTo($li);
+        }
+      }
+    }
+  }
+
+  for (var sampleInfo of this.technicalReplicates) {
```
qiyunzhu: I think this `for ... of` thing is a new feature of the recent ECMAScript 6 standard. May have compatibility issues with old computers?

knimin/static/js/platemapper.js
josenavas: Yes! When updating a well, is triggering the even `change` which will then fire off all the validations.

knimin/static/js/platemapper.js
josenavas: Yes, I've added a comment for clarity

knimin/static/js/platemapper.js
josenavas: The return value is used by [JQuery's grep function](http://api.jquery.com/jquery.grep/).

knimin/static/js/platemapper.js
josenavas: After a quick search it seems like this has been supported since 2014 - I don't think it should be an issue.

knimin/static/js/platemapper.js
josenavas: Good catch - changed to use the obj's properties

knimin/static/js/platemapper.js
josenavas: Done

knimin/templates/pm_plate_list.html
josenavas: Good catch - actually I removed the magic numbers and added them as a list

qiyunzhu:
All look good!

wasade:
:+1:

Issue #197, opened 2017-04-07
Improve error reporting to the user


josenavas:
See discussion [here](https://github.com/biocore/labadmin/pull/195/files#r110287551)

PR #200, opened 2017-04-09
Study creation

josenavas:
~~Review/merge https://github.com/biocore/labadmin/pull/199 first~~

This adds the page for creating a study, and creates the study in Qiita and the project in Jira

### DIFF
--- a/knimin/lib/qiita_client.py
+++ b/knimin/lib/qiita_client.py
@@ -128,7 +128,12 @@ class QiitaClient(object):
                 kwargs['headers']['Authorization'] = 'Bearer %s' % self._token
                 r = req(url, verify=self._verify, **kwargs)
 
-        return r.status_code, r.json()
+        try:
+            r_json = r.json()
+        except Exception:
+            r_json = None
+
+        return r.status_code, r_json
 
     def get(self, url, **kwargs):
         """Execute a get request against the Qiita server
--- a/knimin/lib/qiita_client.py
+++ b/knimin/lib/qiita_client.py
@@ -62,7 +62,6 @@ class QiitaClient(object):
         self._client_id = client_id
         self._client_secret = client_secret
         self._authenticate_url = "%s/qiita_db/authenticate/" % self._server_url
-        print self._authenticate_url
 
         # Fetch the access token
         self._fetch_token()
--- a/knimin/lib/qiita_client.py
+++ b/knimin/lib/qiita_client.py
@@ -62,6 +62,8 @@ class QiitaClient(object):
         self._client_secret = client_secret
         self._authenticate_url = "%s/qiita_db/authenticate/" % self._server_url
 
+        self._session = requests.Session()
+
         # Fetch the access token
         self._fetch_token()
 
@@ -76,8 +78,8 @@ class QiitaClient(object):
         data = {'client_id': self._client_id,
                 'client_secret': self._client_secret,
                 'grant_type': 'client'}
-        r = requests.post(self._authenticate_url, verify=self._verify,
-                          data=data)
+        r = self._session.post(self._authenticate_url, verify=self._verify,
+                               data=data)
         if r.status_code != 200:
             raise ValueError("Can't authenticate with the Qiita server")
         self._token = r.json()['access_token']
@@ -144,7 +146,7 @@ class QiitaClient(object):
         int, dict
             The status code and the JSON response from the server
         """
-        return self._request_oauth2(requests.get, url, **kwargs)
+        return self._request_oauth2(self._session.get, url, **kwargs)
 
     def post(self, url, **kwargs):
         """Execute a post request against the Qiita server
@@ -161,7 +163,7 @@ class QiitaClient(object):
         int, dict
             The status code and the JSON response from the server
         """
-        return self._request_oauth2(requests.post, url, **kwargs)
+        return self._request_oauth2(self._session.post, url, **kwargs)
 
     def patch(self, url, **kwargs):
         """Executes a patch request against the Qiita server
@@ -178,4 +180,4 @@ class QiitaClient(object):
         int, dict
             The status code and the JSON response from the server
         """
-        return self._request_oauth2(requests.patch, url, **kwargs)
+        return self._request_oauth2(self._session.patch, url, **kwargs)
--- a/knimin/lib/qiita_client.py
+++ b/knimin/lib/qiita_client.py
@@ -1,0 +1,182 @@
+# This code is based on the QiitaClient project
+# https://github.com/qiita-spots/qiita_client
+# We can't directly use the QiitaClient library because it doesn't play nice
+# with the new restAPI due to the returning status codes.
+# Hence, including here the copyright notice from the QiitaClient project
+
+# -----------------------------------------------------------------------------
+# Copyright (c) 2014--, The Qiita Development Team.
+#
+# Distributed under the terms of the BSD 3-clause License.
+#
+# The full license is in the file LICENSE, distributed with this software.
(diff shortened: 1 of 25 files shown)


josenavas:
This is ready for review!

knimin/lib/qiita_client.py
```diff
@@ -0,0 +1,186 @@
...
+    -------
+    get
+    post
+    """
+    def __init__(self, server_url, client_id, client_secret, server_cert=None):
+        self._server_url = server_url
+
+        # The attribute self._verify is used to provide the parameter `verify`
+        # to the get/post requests. According to their documentation (link:
+        # http://docs.python-requests.org/en/latest/user/
```
antgonza: What about minimizing the link with goo.gl to make it fit in a single line?

knimin/lib/qiita_client.py
```diff
@@ -0,0 +1,186 @@
...
+                r_json = None
+
+            if r_json and r_json.get('error_description') == \
+                    'Oauth2 error: token has timed out':
+                # The token expired - get a new one and re-try the request
+                self._fetch_token()
+                kwargs['headers']['Authorization'] = 'Bearer %s' % self._token
+                r = req(url, verify=self._verify, **kwargs)
+
+        try:
```
wasade: It looks like the .json call is repeated, why not just do it once in a single try/except?

knimin/lib/qiita_client.py
```diff
@@ -0,0 +1,186 @@
...
+            The url to access in the server
+        kwargs : dict
+            The request kwargs
+
+        Returns
+        -------
+        int, dict
+            The status code and the JSON response from the server
+        """
+        return self._request_oauth2(requests.get, url, **kwargs)
```
wasade: Would it be possible to use a single session? Eg `requests.Session`?

knimin/lib/qiita_jira_util.py
```diff
@@ -1,41 +1,86 @@
...
-    # study_id will be provided by Qiita, hardcoding it
-    study_id = 10001
-    message = ''
+    # Step 1: Create the study in Qiita
+    # Make sure that the qiita_pi and qiita_lp exist in Qiita
+    # We will assume that if the email is given, then the person needs to be
+    # created in qiita
+    for val in [qiita_pi, qiita_lp]:
+        if 'email' in val:
+            status_code, msg = qiita_client.post('/api/v1/person', data=val)
```
wasade: A person is always created? Iirc,this should return 200 if they already exist as it should be idempotent

wasade: This reads as though a study cannot be created if the pi or lab person already exists in the system

knimin/lib/tests/test_qiita_client.py
```diff
@@ -0,0 +1,74 @@
...
+                   'study_description': 'asdasd',
+                   'owner': 'doesnotexist@foo.bar',
+                   'study_alias': 'blah',
+                   'contacts': {'principal_investigator': [u'PIDude',
+                                                           u'Wash U'],
+                                'lab_person': [u'LabDude',
+                                               u'knight lab']}}
+        status_code, data = self.tester.post('/api/v1/study', data=payload,
+                                             as_json=True)
+        # self.assertEqual(status_code, 403)
```
wasade: Can this be corrected?

knimin/lib/tests/test_qiita_jira_util.py
```diff
@@ -1,24 +1,97 @@
...
+            try:
+                f()
+            except Exception as e:
+                print("Database clean-up failed. Downstream tests might be "
+                      "affected by this! Reason: %s" % format_exc(e))
+
+    @classmethod
+    def tearDownClass(cls):
+        # Reset the qiita DB to make tests independent
+        qiita_client.post("/apitest/reset/")
```
wasade: Did this get merged elsewhere? I'm not seeing it right now in the pr or in labadmin

wasade: It would be very good for this endpoint to only be part of webserver if started with a test configuration...

knimin/tests/test_pm_create_study.py
```diff
@@ -0,0 +1,57 @@
...
+            response.effective_url.endswith('?next=%2Fpm_create_study%2F'))
+
+    def test_get(self):
+        self.mock_login_admin()
+        response = self.get('/pm_create_study/')
+        self.assertEqual(response.code, 200)
+        self.assertIn('<h3>Create new project</h3>', response.body)
+        self.assertIn('LabDude (knight lab)</option>', response.body)
+
+    def test_post(self):
```
wasade: Can there be a non auth post test please?

knimin/tests/test_pm_create_study.py
```diff
@@ -0,0 +1,57 @@
...
+        self.mock_login_admin()
+        data = {'qiita-user': 'demo@microbio.me',
```
wasade: Can there be a creation failure as well? The underlying code is tested but it would be good to assert the result is still appropriately handled for the client

knimin/lib/qiita_jira_util.py
josenavas: If the person already exists the [method returns a 409](https://github.com/biocore/qiita/blob/master/qiita_pet/handlers/rest/study_person.py#L41)

The way that I had it here is that only the name and affiliation is required (just as in Qiita). In case that the user wants to create a new user, then the email is also provided. Does it make sense?

knimin/lib/tests/test_qiita_client.py
josenavas: Done!

knimin/lib/tests/test_qiita_jira_util.py
josenavas: Yup, it has been present in [master](https://github.com/biocore/qiita/blob/master/qiita_pet/webserver.py#L190) for a while and it is only present if the webserver is started with a test configuration.

knimin/tests/test_pm_create_study.py
josenavas: good catch! added

knimin/tests/test_pm_create_study.py
josenavas: I added a test to make sure that it returns a 500, which is the current behavior of the system. I do agree that this needs to change and be improved, and we created an [issue](https://github.com/biocore/labadmin/issues/197) a few days ago so we can fix this globally.

knimin/lib/qiita_client.py
wasade: Just do `self._session = requests.Session()` in the constructor and then do `self._session.get` for GET, etc. I don't understand the duplication concern?

wasade:
I did a review instead of flat comment. Just a follow up on the session. I think +1 otherwise

knimin/lib/qiita_client.py
josenavas: Thanks for the info! I've modified it - I though it was going to be more complex to use sessions.

wasade: `requests` is 🍰

knimin/lib/qiita_jira_util.py
wasade: ah okay

wasade:
Build failures btw

wasade:
Still err'ing btw... :(

josenavas:
Still going to fail in that last push - made a typo xD will be pushing again shortly as I detected something else off

josenavas:
@wasade @antgonza the failures are unrelated to the PR (it is the stochastic failure with the geocoder)

PR #204, opened 2017-04-12
Condense dna plates

antgonza:
what it says + pulling from master ....


knimin/lib/data_access.py
```diff
@@ -3802,6 +3802,83 @@ def delete_pool(self, pool_id):
...
+        """
+        with TRN:
+            self._sample_plate_is_unique(name)
+            self._plate_type_exists(plate_type)
+            self._email_exists(email)
+            robot_id = self.get_or_create_property_option_id(
+                "extraction_robot", robot)
+
+            # 2 is 384-well plate
+            if plate_type != 2:
```
ElDeveloper: Can you add tests for these exceptions?

knimin/lib/data_access.py
```diff
@@ -3800,6 +3802,83 @@ def delete_pool(self, pool_id):
...
+                        name, email, created_on, processing_robot_id,
+                        plate_type_id, volume)
+                     VALUES (%s, %s, now(), %s, %s, %s)
+                     RETURNING shotgun_plate_id"""
+            sql_args = [name, email, robot_id, plate_type, volume]
+            TRN.add(sql, sql_args)
+            shotgun_pid = TRN.execute_fetchlast()
+
+            # adding condensed_plates
+            sql = """INSERT INTO pm.condensed_plates
```
qiyunzhu: Is it possible that this SQL query returns error? For example, if the shotgun_plate => dna_plate pair already exists, will it still work?" (I guess yes because the same DNA plate can be prepped multiple times, just double check.)

knimin/lib/data_access.py
```diff
@@ -3800,6 +3802,162 @@ def delete_pool(self, pool_id):
...
+            self._plate_type_exists(plate_type)
+            self._email_exists(email)
+            robot_id = self.get_or_create_property_option_id(
+                "extraction_robot", robot)
+
+            # 2 is 384-well plate
+            if plate_type != 2:
+                raise ValueError("plate_map should be 2: '384-well plate' but "
+                                 "it's %d" % plate_type)
+            if len(dna_plates) != 4:
```
josenavas: Can you modify to check if 0 < len(dna_plates) <= 4 (and fix error message below accordingly?

knimin/lib/data_access.py
```diff
@@ -3800,6 +3802,162 @@ def delete_pool(self, pool_id):
...
+                raise ValueError("You should have 3 plates and you "
+                                 "have %d" % len(dna_plates))
+            wrong_vals = [(pid, pos) for pid, pos in dna_plates
+                          if not self.read_sample_plate(pid) or
+                          (pos < 0 or pos > 3)]
+            if wrong_vals:
+                raise ValueError("There are errors in these "
```
josenavas: `Wrong dna plates position: %s`

PR #206, opened 2017-04-12
WIP: Plate mapper shotgun normalized

wasade:
I do not expect the tests to pass yet (as I didn't have the deps and no internet to get them), but wanted to get visibility on this so far. Are there test plates in place already? Should I be concerned about qpcr details etc?


qiyunzhu:
@wasade @josenavas Is this PR alive and requires code review? I saw that #205 with the same title was open and closed.

wasade:
It is and not yet. Thanks for checking!

josenavas:
@wasade - @antgonza 's PR has been merged so you should be able to write all the tests for your functions 😄  I've changed the title to WIP to we know when its ready for review. Thanks a lot!

wasade:
Excellent, thanks!

wasade:
Okay, I think it's good for a first round of review. There are some inline comments starting with ### which are intended for reviewers to see, and which can be deleted later.

knimin/lib/tests/test_data_access.py
```diff
@@ -1260,6 +1301,142 @@ def test_extract_sample_plates(self):
...
+
+    def test_delete_normalized_shotgun_plate(self):
+        self._create_test_echo()
+        cid = self._create_test_shotgun_plate()
+        nid = db.normalize_shotgun_plate(cid, 'test', 'a valid echo name',
+                                         np.arange(384).reshape(16, 24),
+                                         np.arange(384).reshape(16, 24) * 10)
+        works = db.read_normalized_shotgun_plate(nid)
+        db.delete_normalized_shotgun_plate(nid)
+        try:
```
qiyunzhu: I am bit unclear about the logic here. To my understanding, this plate should have been gone. So what's expected is an error (`pass`). What `else` will be? What if `doesnotwork` worked? Above all, why this test has to be written in this way, instead of the `assertRaisesRegexp` way?

knimin/lib/data_access.py
```diff
@@ -3532,6 +3533,204 @@ def read_dna_plate(self, dna_plate_id):
...
+	Parameters
+	----------
+	shotgun_plate_id : int
+	    The shotgun plate id
+	email : str
+	    The email of the user
+	echo : str
+	    The echo machine performing the normalization
+	plate_normalization_sample : 2D np.array of float
+	    The sample volume in nanoliters for each well
```
josenavas: According to @tanaes notes is nanoliters

knimin/lib/data_access.py
```diff
@@ -3532,6 +3533,204 @@ def read_dna_plate(self, dna_plate_id):
...
+
+            # obtain and verify the echo exists
+            sql = """SELECT echo_id
+                     FROM pm.echo
+                     WHERE echo.name = %s"""
+            TRN.add(sql, [echo])
+            res = TRN.execute_fetchindex()
+            if not res:
+                raise ValueError("echo machine %s does not exist" % echo)
+            else:
```
josenavas: Minor, but I think `else` is not needed, since you raise in the other case.

knimin/lib/data_access.py
```diff
@@ -3532,6 +3533,204 @@ def read_dna_plate(self, dna_plate_id):
...
+                                 "shape %s; expected %s" % (obs, exp))
+
+            # create the instance of the normalized plate
+            sql = """INSERT INTO pm.shotgun_normalized_plate
+                        (shotgun_plate_id, created_on, email, echo_id)
+                     VALUES (%s, now(), %s, %s)
+                     RETURNING shotgun_normalized_plate_id"""
+
+            TRN.add(sql, [shotgun_plate_id, email, echo_id])
+            shotgun_normalized_plate_id = TRN.execute_fetchindex()[0][0]
```
josenavas: Minor, but you can use `TRN.execute_fetchlast()`. is equivalent to `TRN.execute_fetchindex()[0][0]`

knimin/lib/data_access.py
```diff
@@ -3532,6 +3533,204 @@ def read_dna_plate(self, dna_plate_id):
...
+	    The created normalized_shotgun_plate_id
+	"""
+        with TRN:
+            # verify the plate exists
+            sql = """SELECT EXISTS(
+                        SELECT 1
+                        FROM pm.shotgun_plate
+                        WHERE shotgun_plate_id = %s)"""
+            TRN.add(sql, [shotgun_plate_id])
+            res = TRN.execute_fetchindex()[0]
```
josenavas: Minor, but you can use `TRN.execute_fetchlast()`. is equivalent to `TRN.execute_fetchindex()[0][0]`

knimin/lib/data_access.py
```diff
@@ -3532,6 +3533,204 @@ def read_dna_plate(self, dna_plate_id):
...
+        ----------
+        shotgun_normalized_plate_id : int
+            The ID of the normalized shotgun plate to delete.
+
+        Raises
+        ------
+        ValueError
+            If the normalized_shotgun_plate_id does not exist
+        """
+        ### what mechanisms exist to prohibit the execution of this method for
```
josenavas: Currently no deletion method is exposed through the interface.

knimin/lib/data_access.py
```diff
@@ -3532,6 +3533,204 @@ def read_dna_plate(self, dna_plate_id):
...
+        Returns
+        -------
+        dict
+            A dict populated by the columns details in
+            pm.shotgun_normalized_plate, as well as two numpy matrices
+            corresponding to the normalized water and sample volumes
+        """
+        with TRN:
+            # get plate details
+            sql = """SELECT *
```
josenavas: @wasade can you be specific on the request here? Something similar as what is done [here](https://github.com/biocore/labadmin/blob/plate-mapper-kiwi/knimin/lib/data_access.py#L3977).

I know the ID is redundant but we've included it in all the other alike calls, it simplifies a few calls done later.
Also, instead of returning the IDs of the different attributes of the plates, can you return the names? These would be the name of: echo, shotgun_library_prep_kit, shotgun_adapter_aliquot and qpcr

knimin/lib/data_access.py
```diff
@@ -3532,6 +3533,204 @@ def read_dna_plate(self, dna_plate_id):
...
+                     WHERE shotgun_plate_id = %s"""
+            TRN.add(sql, [res['shotgun_plate_id']])
+            plate_shape = dict(TRN.execute_fetchindex()[0])
+            n_rows = plate_shape['rows']
+            n_cols = plate_shape['cols']
+            sample_volumes = np.zeros((n_rows, n_cols))
+            water_volumes = np.zeros((n_rows, n_cols))
+
+            # get well values
+            sql = """SELECT row, col, sample_volume_nl, water_volume_nl
```
josenavas: Can this be extended to include the i5/i7 index names, qpcr_concentration and qpcr cp? Right now the'll be `None` but another method will be adding those values later and we will be using the same method to read those values from the DB

knimin/lib/data_access.py
```diff
@@ -3532,6 +3533,204 @@ def read_dna_plate(self, dna_plate_id):
...
+            sql = """SELECT *
+                     FROM pm.shotgun_normalized_plate
+                     WHERE shotgun_normalized_plate_id = %s"""
+            TRN.add(sql, [shotgun_normalized_plate_id])
+            res = TRN.execute_fetchindex()
+            if not res:
+                raise ValueError("normalized shotgun plate %s does not exist" %
+                                 shotgun_normalized_plate_id)
+
+            sql = """DELETE FROM pm.shotgun_normalized_plate_well_values
```
josenavas: Minor, but you can avoid making the SQL query in line 3716 by modifying this SQL query to return the shotgun_normalized_plate_id. No need to change if you don't feel like but I wanted to mention it.

knimin/lib/tests/test_data_access.py
```diff
@@ -1260,6 +1301,142 @@ def test_extract_sample_plates(self):
...
+        self._create_test_echo()
+        cid = self._create_test_shotgun_plate()
+        nid = db.normalize_shotgun_plate(cid, 'test', 'a valid echo name',
+                                         np.arange(384).reshape(16, 24),
+                                         np.arange(384).reshape(16, 24) * 10)
+        after = datetime.datetime.now()
+        exp_sample = np.arange(384).reshape(16, 24)
+        exp_water = np.arange(384).reshape(16, 24) * 10
+
+        ### many of these values are ignored at this time as dont know
```
josenavas: You can set them to None, and test the entire `dict`. That is the expected behavior since that information is loaded at 2 different steps of the pipeline.

knimin/lib/tests/test_data_access.py
```diff
@@ -1260,6 +1301,142 @@ def test_extract_sample_plates(self):
...
+                                         np.arange(384).reshape(16, 24),
+                                         np.arange(384).reshape(16, 24) * 10)
+        works = db.read_normalized_shotgun_plate(nid)
+        db.delete_normalized_shotgun_plate(nid)
+        try:
+            doesnotwork = db.read_normalized_shotgun_plate(nid)
+        except:
+            pass
+        else:
+            self.fail()
```
josenavas: Beautiful! I didn't know about the fail call!

knimin/lib/tests/test_data_access.py
```diff
@@ -1260,6 +1301,142 @@ def test_extract_sample_plates(self):
...
+    def test_delete_normalized_shotgun_plate(self):
+        self._create_test_echo()
+        cid = self._create_test_shotgun_plate()
+        nid = db.normalize_shotgun_plate(cid, 'test', 'a valid echo name',
+                                         np.arange(384).reshape(16, 24),
+                                         np.arange(384).reshape(16, 24) * 10)
+        works = db.read_normalized_shotgun_plate(nid)
+        db.delete_normalized_shotgun_plate(nid)
+        try:
+            doesnotwork = db.read_normalized_shotgun_plate(nid)
```
josenavas: I think flake8 will complain about this `doesnotwork =` I think you can just leave the call w/o storing the return

knimin/lib/tests/test_data_access.py
```diff
@@ -1260,6 +1301,142 @@ def test_extract_sample_plates(self):
...
+        except:
+            pass
+        else:
+            self.fail()
+
+    def test_read_normalized_shotgun_plate(self):
```
josenavas: I think you can remove this entire test - we've been using the read on the "creation" call (i.e. [this one](https://github.com/biocore/labadmin/pull/206/files#diff-0d1d77e117ce5a979751aa7fba90e5e5R1364)) as a valid success test for the read method. Similarly for the delete method.

knimin/lib/data_access.py
wasade: Will do

knimin/lib/data_access.py
wasade: Will do

knimin/lib/data_access.py
wasade: Will do

knimin/lib/data_access.py
wasade: Will do

knimin/lib/data_access.py
wasade: Will do

knimin/lib/data_access.py
wasade: Okay

knimin/lib/data_access.py
wasade: Wont it except if the ID doesn't exist?

knimin/lib/tests/test_data_access.py
wasade: Okay

knimin/lib/tests/test_data_access.py
wasade: Good call :)

knimin/lib/tests/test_data_access.py
wasade: Good point

knimin/lib/tests/test_data_access.py
wasade: It's super handy especially when stubbing out tests

knimin/lib/tests/test_data_access.py
wasade: okay

knimin/lib/data_access.py
wasade: not seeing qpcr concentration in the schema, or qpcr cp. those details also seem to be part of `pm.shotgun_normalized_plate`? will add i5/i7

knimin/lib/data_access.py
josenavas: I was also surprised about that but it doesn't. If your restriction doesn't hit any record it just runs without removing anything. I guess it makes sense to be able to run some clean up stuff...

```sql
ag_test=# DELETE FROM pm.run_pool WHERE run_pool_id = 1;
DELETE 0
```

PR #207, opened 2017-04-13
Connect Jira and Qiita

josenavas:
This PR adds the methods that updates Qiita and Jira at different steps of the pipeline.

PR #208, opened 2017-04-13
Condense GUI

josenavas:
~~Review/merge #207 first~~

This adds the GUI for condensing plates:

condense

### DIFF
--- a/knimin/handlers/pm_condense.py
+++ b/knimin/handlers/pm_condense.py
@@ -1,0 +1,30 @@
+# -----------------------------------------------------------------------------
+# Copyright (c) 2014--, The LabAdmin Development Team.
+#
+# Distributed under the terms of the BSD 3-clause License.
+#
+# The full license is in the file LICENSE, distributed with this software.
+# -----------------------------------------------------------------------------
+
+from tornado.web import authenticated
+
+from knimin.handlers.base import BaseHandler
+from knimin.handlers.access_decorators import set_access
+from knimin import db
+
+
+@set_access(['Admin'])
+class PMCondensePlatesHandler(BaseHandler):
+    @authenticated
+    def get(self):
+        dna_plates = db.get_dna_plate_list()
+        for plate in dna_plates:
+            plate['date'] = plate['date'].isoformat()
+
+        self.render("pm_condense.html",
+                    plates=dna_plates,
+                    robots=db.get_property_options("processing_robot"))
+
+    @authenticated
+    def post(self):
+        pass
--- a/knimin/handlers/pm_condense.py
+++ b/knimin/handlers/pm_condense.py
@@ -27,4 +27,21 @@ class PMCondensePlatesHandler(BaseHandler):
 
     @authenticated
     def post(self):
-        pass
+        plate1 = self.get_argument('plate-1')
+        plate2 = self.get_argument('plate-2')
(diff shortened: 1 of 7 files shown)


knimin/handlers/pm_condense.py
```diff
@@ -0,0 +1,47 @@
...
+        dna_plates = db.get_dna_plate_list()
+        for plate in dna_plates:
+            plate['date'] = plate['date'].isoformat()
+
+        self.render("pm_condense.html",
+                    plates=dna_plates,
+                    robots=db.get_property_options("processing_robot"))
+
+    @authenticated
+    def post(self):
```
wasade: is it possible that we'd condense less than 4 plates?

josenavas: Yes, but the argument is sent anyways with an empty string. That is covered on line 41.

josenavas:
@qiyunzhu can you take a look to this PR?

Thanks!

knimin/tests/test_pm_condense.py
```diff
@@ -0,0 +1,85 @@
...
+
+        data = {'plate-1': dna_plates[0], 'plate-2': dna_plates[1],
+                'plate-3': dna_plates[2], 'plate-4': dna_plates[3],
+                'name': 'Test plate', 'robot': 'ROBE', 'volume': 0.22}
+
+        self.mock_login_admin()
+        response = self.post('/pm_condense/', data=data)
+
+        plate_ids = []
+        for match in re.findall("plate=[0-9]*", response.effective_url):
```
qiyunzhu: This is a greedy match for the whole numeric part of `plate001122....`, is that right?

knimin/templates/pm_condense.html
```diff
@@ -0,0 +1,94 @@
...
+        <label class="center-block">Processing robot</label>
+        <select name="robot" class="form-control">
+            {% for robot in robots %}
+              <option value='{{robot['name']}}'>{{robot['name']}}</option>
+            {% end %}
+        </select>
+      </div>
+      <div class="col-xs-6 form-group">
+        <label class="center-block">Volume</label>
+        <input class="form-control" name="volume" type="number" min="0" step="0.001" id='volume' required/>
```
qiyunzhu: Will this step be too small? How about `0.01`?

knimin/tests/test_pm_condense.py
josenavas: Correct - I just want to get the plate id so I can clean the DB

knimin/templates/pm_condense.html
josenavas: I agree that as step is pretty small, but the size of the step is the one that defines the precision. All the values that I've seen for this input contains 3 decimals, so that's why I had to put it 0.001. At the end the user can still type the number

tanaes: Can we have units displayed on this input field label just as a reminder?

I think it should be µL for the EPMotion, right?

wasade: :+1:

wasade: i put my thumb in the wrong place

antgonza:
PR hs been reviewed 3+ times and all comments have been addressed + tests are expecting to fail so merging so we move to the one that solves them.

PR #209, opened 2017-04-13
add quantify_shotgun_plate

antgonza:


knimin/lib/data_access.py
```diff
@@ -4218,6 +4222,24 @@ def read_shotgun_plate(self, plate_id):
...
             res['condensed_plates'] = [eval(v)
                                        for v in eval(res['condensed_plates'])]
+            # Get positions
+            sql = """SELECT row, col, sample_id, dna_concentration
+                     FROM pm.shotgun_plate_layout
+                     WHERE shotgun_plate_id = %s"""
+            TRN.add(sql, [plate_id])
+            values = TRN.execute_fetchindex()
+            # getting rows/cols and discarding them from res
+            rows = res['rows']
```
qiyunzhu: `pop` may be a clearner solution.

knimin/lib/data_access.py
```diff
@@ -4218,6 +4222,24 @@ def read_shotgun_plate(self, plate_id):
...
+            del res['rows']
+            cols = res['cols']
+            del res['cols']
+            res['shotgun_plate_layout'] = [[None for c in range(cols)]
```
qiyunzhu: How about: `... = [[None]*len(cols)]*len(rows)`

knimin/lib/data_access.py
```diff
@@ -4258,6 +4284,84 @@ def delete_shotgun_plate(self, plate_id):
...
+            rp = dict(self.read_plate_type(sgp['plate_type_id']))
+            pc_rows, pc_cols = plate_concentration.shape
+            processing_robot_id = self.get_or_create_property_option_id(
+                "plate_reader", plate_reader)
+
+            if pc_rows != rp['rows'] or pc_cols != rp['cols']:
+                raise ValueError('plate_concentration wrong shape, should '
+                                 'be: (%d, %d) but is: (%d, %d)' % (
+                                    rp['rows'], rp['cols'], pc_rows, pc_cols))
+            # we expect 4 plates, thus 4 empty rows
```
josenavas: Note that the 4 plates is an upper limit, we can have less than 4 - is the code supporting this? I'm having a bit of trouble understanding how it is working.

knimin/lib/data_access.py
```diff
@@ -4258,6 +4284,84 @@ def delete_shotgun_plate(self, plate_id):
...
+            If `plate_concentration` dimensions doesn't match the plate type
+        """
+        with TRN:
+            sgp = self.read_shotgun_plate(shotgun_plate_id)
+            rp = dict(self.read_plate_type(sgp['plate_type_id']))
+            pc_rows, pc_cols = plate_concentration.shape
+            processing_robot_id = self.get_or_create_property_option_id(
+                "plate_reader", plate_reader)
+
+            if pc_rows != rp['rows'] or pc_cols != rp['cols']:
```
qiyunzhu: This "concantenation" is always 96 * 4 = 384 for now, but it's also compatible with any number x * 4 = 4x in the future, is that right?

knimin/lib/tests/test_data_access.py
```diff
@@ -1582,28 +1582,51 @@ def test_delete_pool(self):
...
+        samples = []
+        pids = []
+        # gonna create the 4 color plates
+        for i, color in enumerate(['b', 'r', 'y', 'g']):
+            pid = db.create_sample_plate('Test %s' % color, 1, 'test',
+                                         [study_id])
             self._clean_up_funcs.insert(
                 0, partial(db.delete_sample_plate, pid))
+            layout = []
+            for r in range(8):
```
qiyunzhu: Here it seems that `96` is a constant number.

knimin/lib/data_access.py
```diff
@@ -4218,6 +4222,24 @@ def read_shotgun_plate(self, plate_id):
...
                                        for v in eval(res['condensed_plates'])]
+            # Get positions
+            sql = """SELECT row, col, sample_id, dna_concentration
+                     FROM pm.shotgun_plate_layout
+                     WHERE shotgun_plate_id = %s"""
+            TRN.add(sql, [plate_id])
+            values = TRN.execute_fetchindex()
+            # getting rows/cols and discarding them from res
+            rows = res['rows']
+            del res['rows']
```
wasade: why not just omit them from the query on 4195?

knimin/lib/data_access.py
wasade: for the normalized shotgun stuff, i was using `np.array`, would that be good to do here?

knimin/lib/data_access.py
```diff
@@ -4258,6 +4284,84 @@ def delete_shotgun_plate(self, plate_id):
...
+                orders[order] = self.read_sample_plate_layout(sp_id)
+
+            # now we start creating the merged plate
+            data = []
+            for rid, row in enumerate(zip_longest(*orders, fillvalue=None)):
+                for cid, (a, b, c, d) in enumerate(zip_longest(*row,
+                                                   fillvalue=None)):
+                    new_cid = 2 * cid
+                    new_rid = 2 * rid
+                    data.append([
```
wasade: can this be decomposed into a local method? eg

```python
def set_grid(sample_id, plate_id, concentration, row_offset, col_offset, block_row_offset, block_col_offset, data):
    grid_row = row_offset + block_row_offset
    grid_col = col_offset + block_col_offset
    data.append((plate_id, grid_row, grid_col, sample_id, concentration[grid_row, grid_col]))
```

Then the following lines are:

```python
# set the pattern:
# a b
# c d
set_grid(a['sample_id'], shotgun_plate_id, new_rid, new_cid, 0, 0, data)
set_grid(b['sample_id'], shotgun_plate_id, new_rid, new_cid, 0, 1, data)
set_grid(c['sample_id'], shotgun_plate_id, new_rid, new_cid, 1, 0, data)
set_grid(d['sample_id'], shotgun_plate_id, new_rid, new_cid, 1, 1, data)
```

knimin/lib/data_access.py
antgonza: cause I'm actually using them below to create the data structure, I could put it in another SQL but thought this will be faster ...

knimin/lib/data_access.py
antgonza: Saw that but here I preferred to have all values in a single "matrix" as dicts vs multiple matrices

knimin/lib/data_access.py
antgonza: Adding a tests, should be clearer ... also forgot to add a few ifs that will make this work.

knimin/lib/data_access.py
antgonza: Perhaps, rather leave as is ...

knimin/lib/tests/test_data_access.py
antgonza: Yup, it's a tests ...

qiyunzhu: oh  I see. i didn't notice that it's in the test :)

qiyunzhu:
Looks good now! - Qiyun

knimin/lib/data_access.py
josenavas: I agree in this case since this is consistent with read_sample_plate_layout

PR #210, opened 2017-04-13
Quantify targeted plate

antgonza:


knimin/lib/data_access.py
```diff
@@ -3952,6 +3952,70 @@ def prepare_targeted_libraries(self, plate_links, email, robot, tm300tool,
...
+        -----
+            You always will need to first add raw_concentration as it's an
+            insert and mod_concentration, which is an update
+        """
+        valid_data = ['raw_concentration', 'mod_concentration']
+        if data not in valid_data:
+            raise ValueError(
+                "Not a valid data: %s, should be: %s" % (data, valid_data))
+
+        tp = self.read_targeted_plate(plate_id)
```
josenavas: Minor but can you move this block inside the `TRN` block? They're accessing the DB so it makes sense to be inside the transaction.

josenavas:
@qiyun available for a quick look?

qiyunzhu:
@josenavas Sorry was in another task. I will read now.

knimin/lib/tests/test_data_access.py
```diff
@@ -1614,17 +1614,55 @@ def test_prepare_targeted_libraries(self):
...
         for obs_id, exp in zip(obs_ids, exp):
             obs = db.read_targeted_plate(obs_id)
             self.assertTrue(before <= obs.pop('created_on') <= after)
+            for k in obs:
+                self.assertEqual(obs[k], exp[k])
             self.assertEqual(obs, exp)
 
+        # testing quantify_targeted_plate on only one of the plates
+        pid = obs_ids[0]
+        vals = np.random.rand(8, 12)
```
qiyunzhu: So there is no upper / lower limit for the number entered in this box. Is it?

knimin/lib/tests/test_data_access.py
```diff
@@ -1614,17 +1614,55 @@ def test_prepare_targeted_libraries(self):
...
+        db.quantify_targeted_plate(pid, 'mod_concentration', vals)
+        obs = db.read_targeted_plate(pid)
+        self.assertIsNone(obs['raw_concentration'])
+        self.assertIsNone(obs['mod_concentration'])
+        # this should change only raw_concentration
+        db.quantify_targeted_plate(pid, 'raw_concentration', vals)
+        obs = db.read_targeted_plate(pid)
+        npt.assert_almost_equal(obs['raw_concentration'], vals, decimal=5)
+        self.assertIsNone(obs['mod_concentration'])
+        # now changin mod_concentration
```
qiyunzhu: changin => changing

knimin/lib/data_access.py
```diff
@@ -3989,14 +4055,37 @@ def read_targeted_plate(self, plate_id):
...
+                     FROM pm.targeted_plate_well_values
+                     WHERE targeted_plate_id = %s"""
+            TRN.add(sql, [plate_id])
+            vals = TRN.execute_fetchindex()
+            if vals:
+                res['raw_concentration'] = np.zeros((rows, cols))
+                for (row, col, raw_concentration, mod_concentration) in vals:
+                    res['raw_concentration'][row, col] = raw_concentration
+                    if mod_concentration is not None:
+                        if res['mod_concentration'] is None:
```
qiyunzhu: I think that this `res['mod_concentration']` is always `None` before this line.

knimin/lib/data_access.py
```diff
@@ -3952,6 +3952,71 @@ def prepare_targeted_libraries(self, plate_links, email, robot, tm300tool,
...
+
+        Notes
+        -----
+            You always will need to first add raw_concentration as it's an
+            insert and mod_concentration, which is an update
+        """
+        valid_data = ['raw_concentration', 'mod_concentration']
+        if data not in valid_data:
+            raise ValueError(
+                "Not a valid data: %s, should be: %s" % (data, valid_data))
```
josenavas: Changed

knimin/lib/data_access.py
josenavas: This is inside a for loop, so it will be None only during the first iteration of the for loop.

knimin/lib/tests/test_data_access.py
josenavas: nope. The data comes from a file generated by the plate reader.

knimin/lib/tests/test_data_access.py
josenavas: Done

knimin/lib/data_access.py
qiyunzhu: Oh I see :)

knimin/lib/data_access.py
```diff
@@ -3952,6 +3952,71 @@ def prepare_targeted_libraries(self, plate_links, email, robot, tm300tool,
...
+
+        Raises
+        ------
+        ValueError
+            If data is not one of the valid values, see Parameters
+            If vals shape is not the same than the plate
+
+        Notes
+        -----
+            You always will need to first add raw_concentration as it's an
```
antgonza: Not sure, I would say no ... @josenavas ?

josenavas: I think it is ok because we will already have an exception from the DB directly, so we are covered on that end.

wasade: :shipit:

PR #211, opened 2017-04-14
ENH: format sample sheets

wasade:
Using the Illumina Experiment Manager, I was able to load sheets for:

- target gene on miseq
- target gene on hiseq
- shotgun on miseq
- shotgun on hiseq

### DIFF
--- a/knimin/lib/format.py
+++ b/knimin/lib/format.py
@@ -69,7 +69,7 @@ def format_sample_sheet(instrument_type, labadmin_id,
     "lane" key.
 
     If the run type is Target Gene, then sample details are disregarded
-    with the exception of determining the the lanes.
+    with the exception of determining the lanes.
 
     IF the run type is shotgun, then the following keys are required:
         - sample-id
@@ -160,8 +160,8 @@ def format_sample_sheet(instrument_type, labadmin_id,
     base_sheet = _format_general(run_name, labadmin_id, assay, fwd_cycles,
                                  rev_cycles, pi_name, pi_email,
                                  contact_0_name, contact_0_email,
-                                 contact_1_name, contact_1_email,
-                                 contact_2_name, contact_2_email)
+                                 c1name, c1email,
+                                 c2name, c2email)
 
     full_sheet = "%s%s\n%s\n" % (base_sheet, sample_header,
                                  '\n'.join(sample_details))
--- a/knimin/lib/format.py
+++ b/knimin/lib/format.py
@@ -1,0 +1,273 @@
+import datetime
+
+def write_sample_sheet(output_fp, instrument_type, labadmin_id,
+                       run_name, assay, fwd_cycles, rev_cycles,
+                       pi_name, pi_email,
+                       contact_0_name, contact_0_email,
+                       run_type, sample_information,
+                       contact_1_name=None, contact_1_email=None,
+                       contact_2_name=None, contact_2_email=None):
+    full_sheet = format_sample_sheet(output_fp, instrument_type,
+                                     run_name, assay, fwd_cycles, rev_cycles,
+                                     pi_name, pi_email,
+                                     contact_0_name, contact_0_email,
+                                     run_type, sample_information,
+                                     contact_1_name, contact_1_email,
+                                     contact_2_name, contact_2_email)
+    with open(output_fp, 'w') as f:
+        f.write(full_sheet)
+
+
+def format_sample_sheet(instrument_type, labadmin_id,
+                        run_name, assay, fwd_cycles, rev_cycles,
+                        pi_name, pi_email,
+                        contact_0_name, contact_0_email,
+                        run_type, sample_information,
+                        contact_1_name=None, contact_1_email=None,
+                        contact_2_name=None, contact_2_email=None):
+    """Writes a sample sheet
+
+    Parameters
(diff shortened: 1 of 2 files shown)


knimin/lib/format.py
```diff
@@ -0,0 +1,273 @@
...
+        The email of an additional contact.
+
+    Sample Sheet Note
+    -----------------
+    If the instrument type is a MiSeq, any lane information per-sample will be
+    disregarded. If instrument type is a HiSeq, each sample must include a
+    "lane" key.
+
+    If the run type is Target Gene, then sample details are disregarded
+    with the exception of determining the the lanes.
```
josenavas: the the  -> the

knimin/lib/format.py
```diff
@@ -0,0 +1,273 @@
...
+        else:
+            sample_details = [sample_detail_fmt % {'run_name': run_name}]
+    else:
+        sample_details = [sample_detail_fmt % samp
+                          for samp in sample_information]
+
+    base_sheet = _format_general(run_name, labadmin_id, assay, fwd_cycles,
+                                 rev_cycles, pi_name, pi_email,
+                                 contact_0_name, contact_0_email,
+                                 contact_1_name, contact_1_email,
```
josenavas: I think this should be c1name, c1email, etc...

knimin/lib/format.py
```diff
@@ -0,0 +1,273 @@
...
+        sample_header = DATA_TARGET_GENE_STRUCTURE
+        sample_detail = DATA_TARGET_GENE_SAMPLE_STRUCTURE
+    elif run_type == 'Shotgun':
+        sample_header = DATA_SHOTGUN_STRUCTURE
+        sample_detail = DATA_SHOTGUN_SAMPLE_STRUCTURE
+    else:
+        raise ValueError("%s is not a known run type" %
+                         run_type)
+
+    # if its a miseq, there isn't lane information
```
qiyunzhu: This configuration is so specific. In the future it will inevitably need modification.

knimin/lib/format.py
```diff
@@ -0,0 +1,273 @@
...
+    sample_detail_fmt = sample_prefix + sample_detail + sample_suffix
+
+    if run_type == 'Target Gene':
+        if instrument_type == 'hiseq':
+            lanes = sorted({samp['lane'] for samp in sample_information})
+            sample_details = []
+            for idx, lane in enumerate(lanes):
+                # make a unique run-name on the assumption this is required
+                detail = {'lane': lane, 'run_name': run_name + str(idx)}
+                sample_details.append(sample_detail_fmt % detail)
```
qiyunzhu: I didn't know that one can do `%` in a variable :)

knimin/lib/format.py
```diff
@@ -0,0 +1,273 @@
...
+    }
+
+    optional = {
+        'contact_1_name': contact_1_name,
+        'contact_1_email': contact_1_email,
+        'contact_2_name': contact_2_name,
+        'contact_2_email': contact_2_email
+    }
+
+    for k, v in fmt.items():
```
qiyunzhu: Not important, I just wonder if this long function can be simplified by things like `locals` and `exec` with a for loop.

knimin/lib/format.py
```diff
@@ -0,0 +1,273 @@
...
+    elif run_type == 'Shotgun':
+        sample_header = DATA_SHOTGUN_STRUCTURE
+        sample_detail = DATA_SHOTGUN_SAMPLE_STRUCTURE
+    else:
+        raise ValueError("%s is not a known run type" %
+                         run_type)
+
+    # if its a miseq, there isn't lane information
+    if instrument_type == 'miseq':
+        header_prefix = ''
```
qiyunzhu: Trivial. I guess that one can do `header_prefix  = sample_prefix  = ''`

knimin/lib/format.py
```diff
@@ -0,0 +1,273 @@
...
+    c1email = contact_1_email if contact_1_email is not None else ''
+    c2name = contact_2_name if contact_2_name is not None else ''
+    c2email = contact_2_email if contact_2_email is not None else ''
+
+    validate_contact(pi_name, pi_email)
+    validate_contact(contact_0_name, contact_0_email)
+    validate_contact(contact_1_name, contact_1_email)
+    validate_contact(contact_2_name, contact_2_email)
+
+    if fwd_cycles <= 0 or not isinstance(fwd_cycles, int):
```
qiyunzhu: Is `fwd_cycles` a value that the user can type in the interface? I wonder if it's necessary to validate it in the backend.

wasade: The backend will drive this code

knimin/lib/format.py
wasade: Possibly. We only actually care about a few of the fields in here though, same with the instrument

knimin/lib/format.py
wasade: meh

knimin/lib/format.py
wasade: :)

knimin/lib/format.py
wasade: There are likely introspective tricks to be had but it may also make it more difficult to understand.

PR #212, opened 2017-04-14
Adjust create_sequencing_run

wasade:
I think this does the necessary adjustments


josenavas:
Failures here are expected - I'm modifying the interface in another PR

PR #215, opened 2017-04-14
Prepare shotgun libraries

antgonza:


knimin/lib/data_access.py
```diff
@@ -3697,29 +3699,56 @@ def read_normalized_shotgun_plate(self, shotgun_normalized_plate_id):
...
+                    qpcr_concentrations[row, col] = qpcr_concentration
+                if qpcr_cp is not None:
+                    if qpcr_cps is None:
+                        qpcr_cps = np.zeros((n_rows, n_cols))
+                    qpcr_cps[row, col] = qpcr_cp
+                if si5i is not None:
+                    if shotgun_i5_index is None:
+                        shotgun_i5_index = [[None for c in range(n_cols)]
+                                            for r in range(n_rows)]
+                    shotgun_i5_index[row][col] = si5i
```
wasade: Should these be np.array?

knimin/lib/data_access.py
```diff
@@ -3756,6 +3785,195 @@ def delete_normalized_shotgun_plate(self, shotgun_normalized_plate_id):
...
+            The name of the new aliquot
+        notes : str
+            General notes of this aliquot
+        limit_freeze_thaw_cycles : int
+            Limit freeze thaw cycles for this aliquot
+
+        Returns
+        -------
+        int
+            The id of the resently created shotgun adapter aliquot
```
wasade: The id of the created...

knimin/lib/data_access.py
antgonza: Thought about it but in this case these are strings (the primary keys are varchars) so they are list of list of string.

wasade:
Ya, but that's not an issue for numpy. The IDs from biom are np array and
it's nice for indexing. Doesn't really matter here? But may be nice for
consistency of indexing/slicing, although I don't think we're actually
doing that so it may just be extraneous. The reason I pushed for arrya on
the other values is that it's the natural way to represent, avoids jagged
arrays, get slicing, shape, etc

knimin/lib/data_access.py
antgonza: I can change if you feel strongly about it. My problem is how to create an empty list of strings, not sure if I should create it as '', 0s or nans?

wasade:
Meh, let's defer

PR #219, opened 2017-04-17
Echo parsing

wasade:


wasade:
Thanks :)

knimin/lib/parse.py
```diff
@@ -94,3 +95,84 @@ def parse_qpcr_object(contents):
...
+    int
+        The index of the first empty line following the section, or the index
+        position of the end of the file.
+
+    Raises
+    ------
+    ValueError
+        If the section is not found
+    """
+    start = -1
```
qiyunzhu: I feel that this calculation is overly complicated. Maybe possible to interate the `lines` only once. For example:
```
start = -1
end = len(lines)
started = False
for idx, line in enumerate(lines):
    if line.startswith(label):
        start = idx
        started = True
    elif started and not line:
        end = idx
        break
if start == -1:
    raise ValueError('%s section appears to be missing' % label)
```

knimin/lib/parse.py
```diff
@@ -94,3 +95,84 @@ def parse_qpcr_object(contents):
...
+
+    Raises
+    ------
+    ValueError
+        If the header is missing
+        If the exceptions section is missing
+        If the details section is missing
+    """
+    data = data.splitlines()
+    if not data[0].startswith('Run ID'):
```
qiyunzhu: Is there a real chance that the `data` is not formatted correctly?

knimin/lib/tests/test_parse.py
```diff
@@ -97,6 +101,115 @@ def test_parse_qpcr_object(self):
             "u'Status']")
 
 
+class EchoParserTests(TestCase):
+    def setUp(self):
+        base = os.path.join(os.path.dirname(os.path.abspath(__file__)), 'data')
```
qiyunzhu: I wonder if it's easier to do `from os.path import join, dirname, realpath` then do `join` directly, like other scripts in the same project did.

knimin/lib/parse.py
```diff
@@ -94,3 +95,84 @@ def parse_qpcr_object(contents):
...
+        A DataFrame composed of the [DETAILS] section
+
+    Raises
+    ------
+    ValueError
+        If the header is missing
+        If the exceptions section is missing
+        If the details section is missing
+    """
+    data = data.splitlines()
```
qiyunzhu: If `data` has extra trailing empty new lines, this conversion will leave some empty elements in the end. Maybe  do `data.rstrip('\n')` beforehand?

knimin/lib/parse.py
wasade: I'd argue that is more complex...

knimin/lib/parse.py
wasade: that shouldn't be an issue

knimin/lib/parse.py
wasade: I don't know

knimin/lib/tests/test_parse.py
wasade: no strong feelings

wasade:
Failures look unrelated

knimin/lib/parse.py
qiyunzhu: Why? That saves one loop thru the lines, which is expensive.

wasade: This is not performance bound code, and the total number of iterations is the same as at most, the code will the entire file only once as the forloops do not overlap. The multiple loops also represent a logical separation, and I'd argue is more in line with KISS, where each forloop statement does one specific thing.

qiyunzhu: Oh I see. Thanks for explanation!

PR #221, opened 2017-04-17
Added shotgun pooling functions

tanaes:
Adds functions and tests for calculating qPCR concentrations and pooling libraries.

### DIFF
--- a/knimin/lib/shotgun.py
+++ b/knimin/lib/shotgun.py
@@ -36,36 +36,54 @@ def compute_qpcr_concentration(cp_vals, m = -3.231, b = 12.059,
     return(qpcr_concentration)
 
 
-def compute_shotgun_pooling_values_eqvol(sample_concs, total_vol = 60):
+def compute_shotgun_pooling_values_eqvol(sample_concs, total_vol = 60.0):
     """Computes molar concentration of libraries from qPCR Cp values.
 
     Returns a 2D array of calculated concentrations, in nanomolar units
 
     Parameters
     ----------
-    sample_concs : list of float
+    sample_concs : numpy array of float
         The concentrations calculated via qPCR (nM)
     total_vol : float
-        The total volume to pool (µL)
+        The total volume to pool (uL)
 
     Returns
     -------
     np.array of floats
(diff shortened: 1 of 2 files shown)


knimin/lib/shotgun.py
```diff
@@ -0,0 +1,160 @@
...
+        The total volume to pool (uL)
+
+    Returns
+    -------
+    np.array of floats
+        A 2D array of floats
+    """
+    per_sample_vol = (total_vol / sample_concs.size) * 1000.0
+    
+    sample_vols = np.zeros([sample_concs.shape[0], sample_concs.shape[1]]) +
```
wasade: `sample_vols = np.zeros_like(sample_concs) + per_sample_vol`

wasade: ...non blocking request

knimin/lib/shotgun.py
```diff
@@ -0,0 +1,160 @@
...
+    -------
+    sample_vols: np.array of floats
+        the volumes in nL per each sample pooled
+    """
+
+    if sample_fracs is None:
+        sample_fracs = np.ones(sample_concs.shape) / sample_concs.size
+    
+    # get samples above threshold
+    sample_fracs_pass = sample_fracs.copy()
```
wasade: this will throw an exception if `sample_fracs is None`. In addition to adding logic to resolve this, can a test be added which explicitly passes in `sample_fracs=None` to ensure this branch is exercised?

wasade: Nope, I looked at this wrong.

knimin/lib/shotgun.py
tanaes: Fixed!

knimin/lib/shotgun.py
tanaes: Fixed, although still don't have tests to exercise additional cases here (passing / not passing non-default values)

wasade:
I'm :+1:

tanaes:
@josenavas These tests are passing now, and should be ready to merge.

PR #222, opened 2017-04-17
Adding generate_i5_i7_indexes

antgonza:


knimin/lib/data_access.py
```diff
@@ -3894,6 +3894,94 @@ def delete_shotgun_index_aliquot(self, shotgun_index_aliquot_id):
...
+        Parameters
+        ----------
+        idx_tech : str
+            The index technology we want to use
+        num_samples : int
+            The number of pairs to generate
+
+        Returns
+        -------
+        list of (str, str)
```
wasade: This is only true if line 3947 evals True, right?

knimin/lib/data_access.py
```diff
@@ -3894,6 +3894,94 @@ def delete_shotgun_index_aliquot(self, shotgun_index_aliquot_id):
...
+                indices_i7 = [r[0] for r in TRN.execute_fetchindex()]
+                if not indices_i7:
+                    raise ValueError("%s doesn't have any i7 "
+                                     "values" % idx_tech)
+
+                indices = list(product(indices, indices_i7))
+
+            indices_generated = []
+            len_indices = len(indices)
+            # note that this assumes that we will never generate more samples
```
wasade: should an explicit assertion/test be added?

knimin/lib/data_access.py
```diff
@@ -3894,6 +3894,94 @@ def delete_shotgun_index_aliquot(self, shotgun_index_aliquot_id):
...
+            while len(indices_generated) < num_samples:
+                idx_end = idx_start + num_samples
+
+                if idx_end > len_indices:
+                    idx_end = len_indices
+                indices_generated.extend(indices[idx_start:idx_end])
+
+                # in case we have another loop we need to start in 0 and
+                # reset the num_samples requested
+                idx_start = 0
```
wasade: `idx_start` will always be zero with the exception of the first time through the loop, is that correct?

knimin/lib/data_access.py
antgonza: Correct, will change.

knimin/lib/data_access.py
antgonza: I can add.

knimin/lib/data_access.py
antgonza: correct, first time it will be `idx_tech_vals['last_index_idx']:`, line 3965.

knimin/lib/data_access.py
```diff
@@ -3894,6 +3894,103 @@ def delete_shotgun_index_aliquot(self, shotgun_index_aliquot_id):
...
+                raise ValueError('Not a valid shotgun index '
+                                 'technology: %s' % idx_tech)
+            else:
+                # just taking the first element and converting it to dict
+                idx_tech_vals = dict([v for v in idx_tech_vals][0])
+
+            # selecting single indexing
+            sql = """SELECT shotgun_index_id
+                     FROM pm.shotgun_index
+                     WHERE shotgun_index_tech_id = %s AND index_type = 'i5'
```
tanaes: Single index technologies are just i7, so this should be reversed -- only add i5 in dual index strategies.

knimin/lib/tests/test_data_access.py
```diff
@@ -1451,6 +1457,79 @@ def _basic_test_steps_for_normalized_shotgun_plate(self, before, after,
 
             self.assertEqual(obs[k], exp_rnsp[k])
 
+    def test_generate_i5_i7_indexes(self):
+        # these tests will depend on what's in the db, which are added
+        # just below -- Add shotgun_index valid values
+
+        # generate 10 and check the values are just fine
+        obs = db.generate_i5_i7_indexes('iTru', 10)
+        exp = [('iTru5_01_A', 'iTru7_101_01'), ('iTru5_01_A', 'iTru7_101_02'),
```
tanaes: We will want to increment both the i5 and the i7 index with each sample. This has to do with the algorithm the sequencers use for detecting clusters, which depends on a certain amount of base heterogeneity.

knimin/lib/tests/test_data_access.py
```diff
@@ -1451,6 +1457,79 @@ def _basic_test_steps_for_normalized_shotgun_plate(self, before, after,
...
+        with self.assertRaises(ValueError) as ctx:
+            db.generate_i5_i7_indexes('BiooNEXTflex-HT', 10)
+        self.assertEqual(
+            ctx.exception.message, "BiooNEXTflex-HT doesn't have any i5 "
+            "values")
+
+        with self.assertRaises(ValueError) as ctx:
+            db.generate_i5_i7_indexes('Nextera', 10)
+        self.assertEqual(
+            ctx.exception.message, "Nextera doesn't have any i7 values")
```
tanaes: Nextera should have both i5 and i7 -- is this an error just because of the test database?

knimin/lib/tests/test_data_access.py
antgonza: will fix.

knimin/lib/tests/test_data_access.py
antgonza: yup ...

knimin/lib/data_access.py
```diff
@@ -3940,54 +3936,47 @@ def generate_i5_i7_indexes(self, idx_tech, num_samples):
...
+                else:
+                    indices_generated.append((indices_i7[index_num]))
 
-                # in case we have another loop we need to start in 0 and
-                # reset the num_samples requested
-                idx_start = 0
-                num_samples -= len(indices_generated)
+            # this is just to avoid reaching to the larges int PgSQL can store
+            if sample_number >= 2000000000:
+                sample_number = 0
```
wasade: lets say this scenario happens, reuse of the sample number wont cause issues?

antgonza: none ... discussed with @tanaes and he agrees ...

wasade: okay

tanaes: It's really just a counter. The worst thing that could happen would be that you'd start reusing index combinations a bit prematurely.

knimin/lib/data_access.py
```diff
@@ -3894,6 +3894,95 @@ def delete_shotgun_index_aliquot(self, shotgun_index_aliquot_id):
...
+            idx_start = idx_tech_vals['last_index_idx']
+            indices_generated = []
+            for i in range(1, num_samples + 1):
+                sample_number = idx_start + i
+                index_num = (sample_number % len_indices +
+                             (sample_number / len_indices) % len_indices)
+
+                if idx_tech_vals['dual_index']:
+                    indices_generated.append((indices_i5[index_num],
+                                              indices_i7[index_num]))
```
tanaes: I did some more thinking about this, and I have a different suggested solution:

```python
def pick_barcodes(i5, i7, num_samples, start=0):
    samples = {}
    len_i5 = len(i5)
    len_i7 = len(i7)
    for i in range(start,start+num_samples):
        sample_i5 = i % len_i5
        sample_i7 = int(((i % len_i5) + floor(float(i) / float(len_i5))) % len_i7)
        samples[i] = (i5[sample_i5], i7[sample_i7])
    
    return(samples)
```

I think that this gives the correct logic despite different numbers of i5 and i7 barcodes:

```python
In [9]: nextera_i5 = ['D501','D502','D503','D504','D505','D506','D507','D508']

In [10]: nextera_i7 = ['D701','D702','D703','D704','D705','D706','D707','D708','D709','D710','D711','D712']

In [11]: combos = pick_barcodes(nextera_i5, nextera_i7, 96)

In [12]: len(set(combos.values()))
Out[12]: 96

In [13]: combos = pick_barcodes(nextera_i5, nextera_i7, 97)

In [14]: len(set(combos.values()))
Out[14]: 96
```

This scales to larger numbers of barcodes:
```python
In [15]: iTru_i5 = range(48)

In [16]: iTru_i7 = range(48)

In [17]: combos = pick_barcodes(iTru_i5, iTru_i7, 97)

In [18]: len(set(combos.values()))
Out[18]: 97

In [19]: combos = pick_barcodes(iTru_i5, iTru_i7, 48*48)

In [20]: len(set(combos.values()))
Out[20]: 2304

In [21]: combos = pick_barcodes(iTru_i5, iTru_i7, 48*48 + 1)

In [22]: len(set(combos.values()))
Out[22]: 2304
```

The original modulo math I quoted you is wrong I think:
```python
def orig_strategy(i5, i7, num_samples, start=0):
    len_indices = min([len(i7), len(i5)])
    
    samples = {}
    indices_generated = []
    for i in range(start,start+num_samples):
        index_num = (i % len_indices +
                    (i / len_indices) % len_indices)    
        samples[i] = (i5[index_num], i7[index_num])
    
    return(samples)
```

It reuses barcode combos too quickly:
```python
In [24]: combos = orig_strategy(nextera_i5, nextera_i7, 8)

In [25]: len(set(combos.values()))
Out[25]: 8

In [26]: combos = orig_strategy(nextera_i5, nextera_i7, 12)

In [27]: len(set(combos.values()))
Out[27]: 8
```

and it breaks:
```python
In [28]: combos = orig_strategy(nextera_i5, nextera_i7, 16)
---------------------------------------------------------------------------
IndexError                                Traceback (most recent call last)
<ipython-input-28-8429db63caea> in <module>()
----> 1 combos = orig_strategy(nextera_i5, nextera_i7, 16)

<ipython-input-23-a7f3be7d424f> in orig_strategy(i5, i7, num_samples, start)
      7         index_num = (i % len_indices +
      8                     (i / len_indices) % len_indices)    
----> 9         samples[i] = (i5[index_num], i7[index_num])
     10 
     11     return(samples)

IndexError: list index out of range
```

knimin/lib/tests/test_data_access.py
```diff
@@ -1451,6 +1462,60 @@ def _basic_test_steps_for_normalized_shotgun_plate(self, before, after,
...
+
+        # ask for another 10 and make sure it starts where we need to
+        obs = db.generate_i5_i7_indexes('iTru', 10)
+        exp = [('iTru5_01_C', 'iTru7_101_03'), ('iTru5_01_D', 'iTru7_101_04'),
+               ('iTru5_01_E', 'iTru7_101_05'), ('iTru5_01_F', 'iTru7_101_06'),
+               ('iTru5_01_G', 'iTru7_101_07'), ('iTru5_01_H', 'iTru7_101_08'),
+               ('iTru5_02_A', 'iTru7_101_09'), ('iTru5_02_B', 'iTru7_101_10'),
+               ('iTru5_02_C', 'iTru7_101_11'), ('iTru5_02_D', 'iTru7_101_12')]
+        self.assertEqual(obs, exp)
+
```
tanaes: Maybe per comment above, add some tests that verify that the number of actual unique barcode combinations returned is = to the expected maximum set size?

josenavas:
@tanaes do we have the layout of those indexes on their plates? and it is always the same or it changes? we need to store that information to be able to generate the echo pick list.

tanaes:
@josenavas We do have the layout of those indexes on their plates. It should always be the same. 

One wrinkle here is that some barcode types (e.g. nextera) the i5 and i7 are pre-mixed already in the plate, so you can't add them independently. 

Others (e.g. iTru) can be added independently (this is what allows us to practically scale to tens or hundreds of thousands of unique barcodes), with i5 and i7 stored in separate plates. 

Maybe we want to use a barcode_plate concept? I imagine that this could be done after barcode assignment: given the set of barcodes, choose a barcode_plate or _plates that contains those barcodes and then generate the pick list. 

One more wrinkle (and maybe not worth worry about now) is that the modulo / float operation above works really well with the Echo or manually with single indexes but doesn't work well manually or in the EP-motion with pre-mixed double indexes, where use of a multi-channel pipette enforces that column-wise relationships be preserved between the barcode plate and the sample/library plate. One potential resolution for this would be to have an option to have an alternative barcode assignment that just 'stamps' the barcode plate wells onto the corresponding library plate wells.

knimin/lib/data_access.py
```diff
@@ -3894,6 +3894,68 @@ def delete_shotgun_index_aliquot(self, shotgun_index_aliquot_id):
...
+
+                if len_indices <= idx:
+                    idx = 0
+
+            sql = """UPDATE pm.shotgun_index_tech
+                     SET last_index_idx = %s
+                     WHERE shotgun_index_tech_id = %s"""
+            TRN.add(sql, [idx, idx_tech_vals['shotgun_index_tech_id']])
+
+            return indices_generated
```
tanaes: This is cool -- so really by storing the each combo as a separate entry in the DB, we don't have to have separate code for single and dual index? 👍

knimin/lib/data_access.py
```diff
@@ -3894,6 +3894,68 @@ def delete_shotgun_index_aliquot(self, shotgun_index_aliquot_id):
...
+            idx_tech_vals = TRN.execute_fetchindex()
+
+            if not idx_tech_vals:
+                raise ValueError('Not a valid shotgun index '
+                                 'technology: %s' % idx_tech)
+            else:
+                # just taking the first element and converting it to dict
+                idx_tech_vals = dict([v for v in idx_tech_vals][0])
+
+            # selecting the dual section
```
tanaes: I don't quite understand this comment

antgonza: leftover from previous iteration, will remove.

knimin/lib/data_access.py
josenavas: Is there a method to read the information attached to an index? I.e. a method that given the shotgun_index_id it returns all the information about that index? If not, can it be added as it is needed to generate some files down the road?

knimin/lib/data_access.py
```diff
@@ -3894,6 +3894,102 @@ def delete_shotgun_index_aliquot(self, shotgun_index_aliquot_id):
...
+                        last_index_idx
+                     FROM pm.shotgun_index_tech WHERE name = %s"""
+            TRN.add(sql, [idx_tech])
+            idx_tech_vals = TRN.execute_fetchindex()
+
+            if not idx_tech_vals:
+                raise ValueError('Not a valid shotgun index '
+                                 'technology: %s' % idx_tech)
+            else:
+                # just taking the first element and converting it to dict
```
wasade: but why?

knimin/lib/tests/test_data_access.py
```diff
@@ -1451,6 +1457,60 @@ def _basic_test_steps_for_normalized_shotgun_plate(self, before, after,
...
+               'i5_i7_sameplate': False, 'last_index_idx': 0,
+               'dual_index': True, 'i5_name': 'iTru5_01_F', 'i7_col': 21,
+               'i5_row': 5, 'i5_bases': 'GTGAGACT', 'i5_col': 0}
+        self.assertEqual(obs, exp)
+
+        # testing errors
+
+        with self.assertRaises(ValueError) as ctx:
+            db.get_shotgun_index_information(100000000000000)
+        self.assertEqual(ctx.exception.message,
```
wasade: Just a general FYI, using the context variable is a little bit awkward as a context describes a scope and so this is relying on the context "leaking" the variable. I'm not requesting a change, just noting that this is similar to using a leaked forloop variable.

josenavas: Thanks for this info! This is used in multiple places, so we should probably open an issue and change later for `self.assertRaisesRegex`, correct?

antgonza:
Note that errors are not due to the code changes ...

josenavas:
@wasade are your comments addressed?

wasade:
sure

PR #223, opened 2017-04-17
Adding sample sync from qiita

josenavas:
Make sure that the samples are sync'ed with Qiita before opening the plate mapper page.


josenavas:
@wasade small issue that I found when testing with the wet lab - the samples where not being sync'd with Qiita.

wasade:
kk, :+1:

PR #227, opened 2017-04-18
Changing default values

josenavas:
Changing a couple of values as per the wet lab team request.


wasade:
+1. Can't merge from phone until tests pass, and I do not believe this pr impacts any tests

PR #228, opened 2017-04-18
Adding the blanks to the mapper

josenavas:
During the tests with the wet lab we noticed that the blanks where not showing on the dropdown list.

Adding them here.

### DIFF
--- a/knimin/handlers/pm_create_study.py
+++ b/knimin/handlers/pm_create_study.py
@@ -8,11 +8,12 @@
 
 from tornado.web import authenticated
 from tornado.escape import json_decode, json_encode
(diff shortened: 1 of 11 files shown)


wasade:
The test failures appear to be relevant

wasade:
Failures look real

tanaes:
I think this test is failing due to a change in the DB in https://github.com/biocore/american-gut-web/pull/667/files#diff-4a23f2583bf5ae8800eae5c5c77272afL23075 and should pass when merged with the updated test code for `test_normalize_shotgun_plate`.

josenavas:
💥 Can somebody merge?

PR #230, opened 2017-04-18
Adding name to some methods

antgonza:
~This is _not_ ready for review~

### DIFF
--- a/knimin/lib/data_access.py
+++ b/knimin/lib/data_access.py
@@ -2458,6 +2458,19 @@ class KniminAccess(object):
 
     # - PlateMapper functions - #
 
+    def get_blanks(self):
+        """Retrieves the list of samples that are blanks
+
+        Returns
+        -------
+        list of str
+            The sample ids of the blank samples
+        """
+        with TRN:
+            sql = "SELECT sample_id FROM pm.sample WHERE is_blank = true"
+            TRN.add(sql)
+            return TRN.execute_fetchflatten()
+
     def get_studies(self):
         """Retrieves the list of studies present in the DB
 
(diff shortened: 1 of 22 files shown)


josenavas:
This is ready for review

antgonza:
Note that @josenavas and me pair programmed this.

josenavas:
@wasade are you available for a review?

knimin/lib/data_access.py
```diff
@@ -3477,6 +3477,8 @@ def extract_sample_plates(self, sample_plate_ids, email, robot, kit_lot,
             The name of the tool used for extraction
         notes : str, optional
             Notes added to the extracted plates
+        names : lilst of str, optional
```
wasade: lilst -> list

knimin/lib/data_access.py
```diff
@@ -3496,19 +3498,29 @@ def extract_sample_plates(self, sample_plate_ids, email, robot, kit_lot,
...
+                            email, name, created_on, sample_plate_id,
+                            extraction_robot_id, extraction_kit_lot_id,
+                            extraction_tool_id, notes)
+                         VALUES (%s, %s, now(), %s, %s, %s, %s, %s)
+                         RETURNING dna_plate_id"""
             dna_plates = []
-            for p_id in sample_plate_ids:
-                TRN.add(sql, [email, p_id, p_id, robot_id, kit_lot_id,
-                              tool_id, notes])
+            for p_id, name in zip_longest(sample_plate_ids, names,
```
wasade: if `len(names) > len(sample_plate_ids)`, then a `None` will be used as the `p_id` placing a `null` in the database for the `sample_plate_id`. Is that correct? It looks like the following should be enforced:

```python
if names:
    if len(names) != len(sample_plate_ids):
        raise ValueError("...")
```

knimin/lib/data_access.py
```diff
@@ -3460,7 +3460,7 @@ def get_sample_plate_list(self):
             return plates
 
     def extract_sample_plates(self, sample_plate_ids, email, robot, kit_lot,
-                              tool, notes=None):
+                              tool, notes=None, names=[]):
```
wasade: it is dangerous to use a default like this, please change to `names=None`, test for it, and set the default within the method

ElDeveloper: 👍

knimin/lib/tests/test_data_access.py
```diff
@@ -1776,6 +1783,43 @@ def test_prepare_targeted_libraries(self):
...
+             'dna_plate_id': dna_plate_ids[1], 'primer_plate_id': 2,
+             'master_mix_lot': '14459', 'robot': 'ROBE',
+             'tm300_8_tool': '208484Z', 'tm50_8_tool': '108364Z',
+             'raw_concentration': None, 'mod_concentration': None,
+             'water_lot': 'RNBD9959'}]
+        for obs_id, exp in zip(obs_ids, exp):
+            obs = db.read_targeted_plate(obs_id)
+            self.assertTrue(before <= obs.pop('created_on') <= after)
+            for k in obs:
+                self.assertEqual(obs[k], exp[k])
```
ElDeveloper: Why is this per-key comparison needed if the following test is asserting the equality of the two objects as a whole?

knimin/lib/data_access.py
```diff
@@ -4953,24 +4969,28 @@ def create_sequencing_run(self, pool_id, email, sequencer, reagent_type,
             The number of forward cycles used.
         rev_cycles : int
             The number of reverse cycles used.
+        name : str, optional
+            Name to add, if provided
```
ElDeveloper: Can you note that if not provided, the default name will be as given by the pool i.e. what's happening in line 4982.

antgonza:
@wasade and @ElDeveloper, could you take another look? Thanks!

knimin/lib/data_access.py
```diff
@@ -3516,9 +3516,18 @@ def extract_sample_plates(self, sample_plate_ids, email, robot, kit_lot,
                          VALUES (%s, %s, now(), %s, %s, %s, %s, %s)
                          RETURNING dna_plate_id"""
             dna_plates = []
+
+            if names:
+                if len(names) != len(sample_plate_ids):
+                    raise ValueError(
```
ElDeveloper: Can you add a note about this in the `Raises` section?

ElDeveloper:
Looks good to me, @wasade can you merge this whenever you get a chance?

ElDeveloper:
thanks @antgonza!

PR #231, opened 2017-04-18
update AG metadata

wasade:
@josenavas, I can't test this locally. Will msg out out-of-band to coordinate

### DIFF
--- a/knimin/handlers/barcode_util.py
+++ b/knimin/handlers/barcode_util.py
@@ -344,7 +344,7 @@ def update_ag_metadata(barcode, md):
         # already represented; this status code means that the sample is
         # already present in the qiita study. there is a race condition
         # though as we cannot put a lock on the remote resouce. So it is
-        # techinically possible that a sample metadata for this sample will
+        # technically possible that a sample metadata for this sample will
         # get loaded via a different mechanism in between the time we check
         # qiita, and the time we push sample metadata.
 
--- a/knimin/handlers/barcode_util.py
+++ b/knimin/handlers/barcode_util.py
@@ -12,7 +12,7 @@ from knimin.handlers.access_decorators import set_access
 class BarcodeUtilHelper(object):
     def get_ag_details(self, barcode):
         ag_details = db.getAGBarcodeDetails(barcode)
-        _, failures = db.pulldown([barcode], [])
+        md, failures = db.pulldown([barcode], [])
 
         if len(ag_details) == 0 and failures:
             div_id = "no_metadata"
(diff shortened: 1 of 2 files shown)


josenavas:
@wasade the failures here look legit - mainly import errors though

knimin/handlers/barcode_util.py
```diff
@@ -305,3 +312,48 @@ def post(self):
...
+                                      as_json=True)
+
+    if sc == 201:
+        return "Metadata for AG sample %s added into Qiita"
+    elif sc == 200:
+        # this should not happen as we first check whether the sample is
+        # already represented; this status code means that the sample is
+        # already present in the qiita study. there is a race condition
+        # though as we cannot put a lock on the remote resouce. So it is
+        # techinically possible that a sample metadata for this sample will
```
qiyunzhu: `techinically` => `technically`

knimin/handlers/barcode_util.py
```diff
@@ -305,3 +312,48 @@ def post(self):
...
+    if sc != 200:
+        return "Unable to get sample IDs from Qiita"
+
+    existing = set(json_decode(response.body))
+    if barcode in existing:
+        return "Metadata for %s already exists in Qiita" % barcode
+
+    # the db.pulldown call returns formatted metadata, so we actually need to
+    # reparse it unfortunately.
+    md = pd.read_csv(md, sep='\t', dtype=str)
```
ElDeveloper: Unfortunately `dtype=str` is not sufficient to preserve string-only types, I suggest that you also add: `na_values=[], keep_default_na=False`.

knimin/handlers/barcode_util.py
```diff
@@ -305,3 +312,48 @@ def post(self):
...
+    md.rename({'#SampleID': 'sample_name'}, inplace=True)
+    sc, response = qiita_client.patch('/api/v1/study/10317',
+                                      data=md.todict(),
```
ElDeveloper: Is this dictionary correctly created? I don't think the dataframe is indexed, so I don't know what is going to happen here. The safest choice might be to explicitly set the index.

knimin/handlers/barcode_util.py
wasade: really??? damn. okay. very good to know. I'm assuming this is situations where its like "\t\t"?

ElDeveloper: That and NA, that's why keep_default_na is set to False.

josenavas:
@wasade I think there still legit failures in here

PR #232, opened 2017-04-19
Echo shotgun pool

josenavas:
Generates the echo pick list for pooling

### DIFF
--- a/knimin/handlers/pm_shotgun_pool.py
+++ b/knimin/handlers/pm_shotgun_pool.py
@@ -14,6 +14,7 @@ from knimin import db
 from knimin.handlers.base import BaseHandler
 from knimin.handlers.access_decorators import set_access
 from knimin.lib.parse import parse_qpcr_object
+from knimin.lib.format import format_pooling_echo_pick_list
 from knimin.lib.shotgun import (compute_qpcr_concentration,
                                 compute_shotgun_pooling_values_qpcr)
 
@@ -71,10 +72,3 @@ class PMShotgunPool(BaseHandler):
         self.write(content)
         self.flush()
         self.finish()
-
-
-def format_pooling_echo_pick_list(arg):
-    # FIXME: This needs to be implemented, and appropriately tested
-    if arg is None:
-        pass
-    return ''
--- a/knimin/lib/format.py
+++ b/knimin/lib/format.py
@@ -72,7 +72,7 @@ def format_normalization_echo_pick_list(vol_sample, vol_water):
 
 
 def format_pooling_echo_pick_list(vol_sample):
-    """Format the contents of an echo pooling pick list file_name
+    """Format the contents of an echo pooling pick list
 
     Parameters
     ----------
--- a/knimin/lib/format.py
+++ b/knimin/lib/format.py
@@ -71,6 +71,33 @@ def format_normalization_echo_pick_list(vol_sample, vol_water):
     return "\n".join(contents)
 
 
+def format_pooling_echo_pick_list(vol_sample):
+    """Format the contents of an echo pooling pick list file_name
+
+    Parameters
+    ----------
+    vol_sample : 2d numpy array of floats
+        The per well sample volume
+    """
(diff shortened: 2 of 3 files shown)


knimin/handlers/pm_shotgun_pool.py
```diff
@@ -0,0 +1,74 @@
...
+            qpcr_concentrations, min_conc=min_conc, floor_conc=floor_conc,
+            total_nmol=total_nmol)
+
+        content = format_pooling_echo_pick_list(sample_vols)
+
+        db.qpcr_shotgun(plate_id, self.get_current_user(), qpcr_machine, '',
+                        qpcr_data, qpcr_concentrations)
+
+        # setup a unique filename
+        file_name = "echo_pool_%s_%s_%s.csv" % (
```
qiyunzhu: Is it possible that the wet lab team worked on the same plate twice in a day?

knimin/lib/tests/test_format.py
```diff
@@ -53,6 +54,23 @@ def test_format_normalization_echo_pick_list(self):
...
+        obs = format_pooling_echo_pick_list(vol_sample)
+        obs_lines = obs.splitlines()
+
+        self.assertEqual(
+            obs_lines[0],
+            'Source Plate Name,Source Plate Type,Source Well,Concentration,'
+            'Transfer Volume,Destination Plate Name,Destination Well')
+        self.assertEqual(
+            obs_lines[1],
+            "1,384LDV_AQ_B2_HT,A1,"",1.50,NormalizedDNA,A1")
```
qiyunzhu: What does the `""` in the middle do? Should that be escaped?

knimin/lib/format.py
```diff
@@ -71,6 +71,33 @@ def format_normalization_echo_pick_list(vol_sample, vol_water):
     return "\n".join(contents)
 
 
+def format_pooling_echo_pick_list(vol_sample):
+    """Format the contents of an echo pooling pick list file_name
```
ElDeveloper: This is formatting the contents of a file, can you remove the file_name part.

knimin/lib/format.py
```diff
@@ -71,6 +71,33 @@ def format_normalization_echo_pick_list(vol_sample, vol_water):
...
+                'Destination Well']
+    # Write the sample transfer volumes
+    rows, cols = vol_sample.shape
+    for i in range(rows):
+        for j in range(cols):
+            well_name = "%s%d" % (chr(ord('A') + i), j+1)
+            # Machine will round, so just give it enough info to do the
+            # correct rounding. The desination well is A1 because we're pooling
+            # samples
+            val = "%.2f" % vol_sample[i][j]
```
ElDeveloper: Are two digits sufficient?

tanaes: This should be in nL, and the machine has a resolution of 2.5 nL. So two digits after the dot is great.

Probably a good idea to include units in comments as frequently as possible to avoid issues.

knimin/handlers/pm_shotgun_pool.py
josenavas: I guess that is possible but this will not overwrite any file, since this is just to give a name for the file that the wet lab will download, so I think it is ok.

knimin/lib/tests/test_format.py
josenavas: Actually it doesn't do anything, so I removed them.

knimin/lib/format.py
josenavas: Thanks atom auto-complete! hahaha

josenavas:
Thanks! Commits are gone - I was missing sync-ing with upstream

qiyunzhu:
Looks good now!

ElDeveloper:
Looks good to me, but tests are failing.

josenavas:
Thanks for the heads up @ElDeveloper 
The failing test is a stochastic test that occurs sometimes when connecting to the google API, unrelated to the changes here

ElDeveloper:
Fair enough, since this is only a google api-related failure I am merging this, thanks!

PR #233, opened 2017-04-20
Parse plate reader output multiple

antgonza:


knimin/lib/parse.py
```diff
@@ -176,3 +176,62 @@ def parse_echo(data):
...
+            if line[0] != chr(ord('A') + lines_read):
+                raise ValueError(
+                    'Wrong row value: %c, it should be: %c, on line %d, of '
+                    'frame %d' % (line[0], chr(ord('A') + lines_read),
+                                  lines_read + 1, i + 1))
+
+            for j, v in enumerate(line[1:]):
+                qubit[i][lines_read, j] = float(v)
+
+            fidx += 1
```
qiyunzhu: What's the purpose of this line?

antgonza: Excellent question, I think I left it by mistake ...

antgonza: sorry, that line is actually required. if you see the for loops basically we start with a fidx (where we find a header) but then we incremented to go line by line ... in other words, fidx starts as the first line of the frame and then we use it to go line by line.

PR #247, opened 2018-09-14
Vioscreen sync

AlexanderFengFeng:
Primarily the VioscreenHandler class that should pull external food frequency questionnaire data from the Vioscreen API and format and store data into AG tables for future use. Also includes unittest class.


knimin/lib/vioscreen.py
```diff
@@ -0,0 +1,446 @@
...
+    
+    def get_users(self):
+        """Gets list of users that vioscreen has data for
+
+        Return
+        ------
+        list of dict
+            List of users that have vioscreen data
+        """
+        if not self._headers:
```
wasade: would it make sense for `self._headers` to get set in the constructor?

knimin/lib/vioscreen.py
```diff
@@ -0,0 +1,446 @@
...
+            self._headers = {'Accept': 'application/json',
+                             'Authorization': 'Bearer %s' % self.get_token()}
+        return self.get('https://api.viocare.com/%s/users' % self._key,
+                         headers=self._headers)
+
+    def get(self, url, retries=5, **kwargs):
```
wasade: Would it be possible to get docstrings?

knimin/lib/vioscreen.py
```diff
@@ -0,0 +1,446 @@
...
+        for i in range(retries):
+            req = self._session.get(url, **kwargs)
+            if req.status_code != 200:  # HTTP status code, 200 is all good
+                data = req.json()
+
+                # if we did not get a HTTP status code 200, than guess that the
+                # API token is no longer valid so get a new one and retry
+                if 'Code' in data and data['Code'] == 1016:
+                    self._headers['token'] = self.get_token()
```
wasade: It looks like it's possible for `self._headers` to be `None`. For example:

```python
obj = VioscreenHandler()
foo = obj.get(...)
```

...in which case, the code would probably through a `KeyError`?

knimin/lib/vioscreen.py
```diff
@@ -0,0 +1,446 @@
...
+                else:
+                    print(self._session.get, url, kwargs)
```
wasade: Can the print statements be removed please?

knimin/lib/vioscreen.py
```diff
@@ -0,0 +1,446 @@
...
+            Session ID that data is being requested for
+        endpoint: str
+            Name of the session type that data is being requested for
+
+        Return
+        ------
+        dict:
+            Food frequency questionnaire data
+        """
+        if not self._headers:
```
wasade: It looks like this if-statement is replicated logic, should it be centralized into a boilerplate method? One strategy here would be to make an idempotent function, something like:

```python
class VioscreenHandler(object):
    ...
    def _assure_headers(self):
        if not self._headers:
            self._headers = {'Accept': 'application/json',
                             'Authorization': 'Bearer %s' % self.get_token()}
```

...and then in methods which need the headers, they could always just have the boilerplate call:

```python
class VioscreenHandler(object):
    ...
    def get_session_data(...):
        """ ... """
        self._assure_headers()
        ...
```

knimin/lib/vioscreen.py
```diff
@@ -0,0 +1,446 @@
...
+            self._headers = {'Accept': 'application/json',
+                             'Authorization': 'Bearer %s' % self.get_token()}
+        if not self._users:
+            self._users = self.get_users()
+
+        if user_ids:
+            if type(user_ids) is not set:
+                raise TypeError('user_ids should be type set')
```
wasade: Since we control the signature of `get_users`, can't this constraint be enforced by the method such that it always returns a `set`?

knimin/lib/vioscreen.py
```diff
@@ -0,0 +1,446 @@
...
+        survey_ids = self.get_init_surveys()
+
+        for user in users['users']:
+            username = user['username']
+
+            try:
+                session_data = self.get('https://api.viocare.com/%s/users/%s/sessions'
+                                        % (self._key, username), headers=self._headers)
+            except ValueError:
+                # I don't understand this, but "JDebelius" does not exist.
```
wasade: Weird

I wonder if that username should be filtered out by `get_users`?

knimin/lib/tests/test_vioscreen.py
```diff
@@ -0,0 +1,327 @@
...
+from os.path import join, dirname, realpath
+from six import StringIO
+import json
+
+from knimin.lib.vioscreen import VioscreenHandler
+
+
+class TestVioscreenHandler(TestCase):
+    @classmethod
+    def setUpClass(cls):
```
wasade: Why `cls` instead of `self`?

knimin/lib/vioscreen.py
AlexanderFengFeng: Setting the headers takes the most time. I remember showing you running the script and it taking some time, so I made this change so testing would be expedient if I didn't need the headers. But I can see why I could just set it in the constructor. Would you still want me to go through with the change.

wasade: There's a boilerplate suggestion below too, which could probably be minimally extended to use as a decorator

knimin/lib/vioscreen.py
AlexanderFengFeng: user_ids is referring to the input by the user, which is not explicitly set yet. If the input, user_ids is None, then user_ids will be set to all of the user_ids pullable from VioScreen like {'surveyid1', 'surveyid2', ...., 'surveyidn'}, where if you set user_ids to something like {'surveyid1'}, it'll just pull info for 'surveyid1'. self._users is different in that it contains users with a lot of extraneous data that isn't necessary for this script, so I use user_ids to filter the users that are needed for pulling.

knimin/lib/vioscreen.py
AlexanderFengFeng: I've never seen this error occur in my many tests.

knimin/lib/tests/test_vioscreen.py
AlexanderFengFeng: I looked up setUpClass() which was necessary for my testing because it only sets up once for all of the testing instead of before each test. It's listed as a classmethod (which I haven't exposure to before) and in looking up classmethod, it looks like each use of it has cls as the subject. https://stackoverflow.com/questions/12179271/meaning-of-classmethod-and-staticmethod-for-beginner

knimin/lib/vioscreen.py
AlexanderFengFeng: Hm I just moved the headers and users initialization back to the constructor, which would hypothetically have headers and users set always. Would this suffice?

knimin/lib/tests/test_vioscreen.py
wasade: Oh, I see... ya, okay

.travis.yml
```diff
@@ -1,61 +1,51 @@
...
+    - oracle-java8-set-default
 before_install:
-  - redis-server --version
-  - wget http://repo.continuum.io/miniconda/Miniconda3-latest-Linux-x86_64.sh -O miniconda.sh
-  - chmod +x miniconda.sh
-  - ./miniconda.sh -b
-  - export PATH=/home/travis/miniconda3/bin:$PATH
-  # Update conda itself
-  - conda update --yes conda
+- openssl aes-256-cbc -K $encrypted_295d008f0223_key -iv $encrypted_295d008f0223_iv -in .auth.json.enc -out .auth.json -d
```
wasade: I was assuming we'd use travis secrets?

https://docs.travis-ci.com/user/encryption-keys/

AlexanderFengFeng:
@wasade I went through the steps of encrypting the auth file, and encountered failures multiple times. Apparently, encrypted files are not available for pull requests from forks because of some relation with environment variables ([https://docs.travis-ci.com/user/encrypting-files/](url)). Unfortunately, it doesn't look like I'll be able to get auth file integrated into the pull request on my end, so we'd have to figure out a work around.

Additionally, I've updated my code based on your notes and the notes from code review. So, whenever you have time.

.travis.yml
AlexanderFengFeng: It looks like that may not be possible either because the travis documentation lists **Please note that encrypted environment variables are not available for pull requests from forks.**

wasade:
Closing in favor of #250 which includes this commit history

PR #252, opened 2019-03-15
WIP: Buffer for qiita updates

wasade:
Resolves initial criteria.

Note that no unit tests were added as the database functions are quite thin though could be convinced otherwise. The print statement with the sending message is where I'm intending to slot in @charles-cowart's code. 

The progress notice in the GIF below does not update *yet* as the print statement executes immediately.

buffer_example


knimin/handlers/barcode_util.py
```diff
@@ -172,8 +173,35 @@ def _build_email(self, login_user, barcode, email_type,
...
+    @authenticated
+    def get(self):
+        barcodes = db.get_unsent_barcodes_from_qiita_buffer()
+        status = db.get_send_qiita_buffer_status()
+        dat = {'status': status, "barcodes": barcodes}
+        self.write(json_encode(dat))
+        self.finish()
+
+    @authenticated
+    def post(self):
```
antgonza: Do we need some kind of check to know that there is no other browser sending already?

knimin/lib/data_access.py
```diff
@@ -1402,6 +1402,60 @@ def getAGKitDetails(self, supplied_kit_id):
...
+        """Adds barcode to the qiita buffer
+
+        Parameters
+        ----------
+        barcode : str
+            The identifier to stage
+        """
+        sql = """SELECT barcode
+                 FROM project_qiita_buffer"""
+        present = {i[0] for i in self._con.execute_fetchall(sql)}
```
antgonza: not sure if worth considering but this will fail if there are no barcodes ...

antgonza: perhaps worth replacing for fetchflatten, but not sure if available in code base

knimin/lib/data_access.py
```diff
@@ -1402,6 +1402,60 @@ def getAGKitDetails(self, supplied_kit_id):
...
+        else:
+            sql = """INSERT INTO project_qiita_buffer (barcode)
+                     VALUES (%s)"""
+            self._con.execute(sql, [barcode])
+            return "Barcode inserted"
+
+    def get_send_qiita_buffer_status(self):
+        """Obtain the present status of the Qiita submission buffer"""
+        sql = """SELECT state FROM project_qiita_buffer_status"""
+        return self._con.execute_fetchone(sql)
```
antgonza: This assumes that there is one and only one, is that OK?

knimin/lib/data_access.py
```diff
@@ -1402,6 +1402,60 @@ def getAGKitDetails(self, supplied_kit_id):
...
+                 SET state = %s
+                 WHERE id = 0"""
+        self._con.execute(sql, [state])
+
+    def get_unsent_barcodes_from_qiita_buffer(self):
+        """Extract the barcodes that have not been sent to Qiita"""
+        sql = """SELECT barcode
+                 FROM project_qiita_buffer
+                 WHERE pushed_to_qiita='N'"""
+        return [i[0] for i in self._con.execute_fetchall(sql)]
```
antgonza: This also assumes that there is always 1 ... safe?

knimin/lib/data_access.py
wasade: Sorry, wrong button.  Not sure if fetch flatten is in here, will use if available, thanks!

knimin/lib/data_access.py
wasade: Shouldn't as it only indexes if there are rows

knimin/handlers/barcode_util.py
wasade: Unclear. Will make a note, but doesn't prohibit testing the larger interaction so would prefer to not resolve this question within this pull request

knimin/lib/data_access.py
wasade: yes

wasade:
I'll put tests in for the database interaction methods

knimin/handlers/barcode_util.py
```diff
@@ -1,12 +1,42 @@
...
+from qiita_client import QiitaClient
 
 from knimin import db
 from knimin.lib.constants import survey_type
 from knimin.lib.mail import send_email
 from knimin.handlers.access_decorators import set_access
+from knimin.lib.configuration import config
+
+
+def get_qiita_client():
```
antgonza: I couldn't find a test for this.

knimin/handlers/barcode_util.py
```diff
@@ -172,8 +202,61 @@ def _build_email(self, login_user, barcode, email_type,
...
+def align_with_qiita_categories(samples, categories):
+    # obviously should instead align to the actual sample metadata...
+    aligned = {}
+    for s in samples:
+        aligned[s] = {c: 'LabControl test' for c in categories}
+    return aligned
+
+
+@set_access(['Scan Barcodes'])
+class PushQiitaHandler(BaseHandler):
```
antgonza: Are there tests for this class?

knimin/handlers/barcode_util.py
```diff
@@ -172,8 +202,61 @@ def _build_email(self, login_user, barcode, email_type,
...
+    executor = concurrent.futures.ThreadPoolExecutor(5)
+    study_id = config.qiita_study_id
+    qclient = get_qiita_client()
+
+    @concurrent.run_on_executor
+    def _push_to_qiita(self, study_id, samples):
+        # TODO: add a mutex or block to ensure a single call process at a time
```
antgonza: Should we have this as an issue?

knimin/handlers/barcode_util.py
```diff
@@ -172,8 +202,61 @@ def _build_email(self, login_user, barcode, email_type,
...
+        cats = self.qclient.get('/api/v1/study/%s/samples/info' % study_id)
```
antgonza: Should `/api/v1/study/` be a class variable in case it changes?

knimin/handlers/barcode_util.py
antgonza: Thanks, add issue so we can have resolution?

knimin/handlers/barcode_util.py
wasade: Yes the added handler tests

knimin/handlers/barcode_util.py
wasade: I guess the benefit isn't really clear to me for this method. But can add something to exercise it.

knimin/handlers/barcode_util.py
wasade: if you want to add. TODOs are typical in some of our codebases too. The likelihood of multiple executions here is low, and even then, sending the same sample multiple times to qiita is allowed as it's treated like an update

knimin/handlers/barcode_util.py
wasade: It also obfuscates the endpoint. These haven't changed in 2 years and I don't believe there is a plan for a v2?

knimin/handlers/barcode_util.py
wasade: See the mutex TODO above

knimin/handlers/barcode_util.py
antgonza: Note that depending on the scenario an update will take hours vs. minutes as Qiita, currently, will need to update the sample in all the preps where is present. Now, in my experiences, TODOs in the code base never get revisited cause they are not easy to find; while issues do; thus, I try to open an issue for each of the TODOs. BTW there is a cool feature in github that allows you to create an issue from any of these conversations. New issue: https://github.com/biocore/labadmin/issues/253

knimin/handlers/barcode_util.py
antgonza: IMOO the benefit is that you can check that the method actually connects to Qiita; which is missing from the added tests. Is this intentional?

knimin/handlers/barcode_util.py
antgonza: Thanks!

knimin/handlers/barcode_util.py
antgonza: Not sure ... let's leave as is.

knimin/handlers/barcode_util.py
wasade: Yes, as it would require either storing credentials in travis which is annoying or greatly complicate the CI environment. This is executed at process start, so if `QiitaClient` doesn't initiate, it will be blazingly obvious too.

wasade:
Thanks!!d

PR #254, opened 2019-03-19
Sample push exception capturing

charles-cowart:
Added logging to PushQiitaHandler. This should resolve the immediate need, but it could be done across the board with a little bit of time. Logging is being reported on screen and in the log file. Curiously, I did not encounter an exception while testing the code, and all pushes were successful. I will issue a second PR to catch a broader range of exceptions.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -1,7 +1,10 @@
 Knight Lab administrative portal
 ================================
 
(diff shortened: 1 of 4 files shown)


charles-cowart:
It looks like in the Travis build, the pushed Ids are not being set to 'Y' afterward. A little strange because it's passing in my local build. Looking into it.

wasade:
Superseded by #255

